PR #1, opened 2022-11-14
Add ADR folder

fabriziosestito:
Add ADR folder with some ADR about checks, some guidance about the tooling, and the `.tool-version` file


dottorblaster (approved):
LGTM

PR #16, opened 2023-02-08
Run CI step `test` on multiple versions of Elixir & OTP

jamie-suse:
This change aims to run the `test` step of the CI against both the versions of Elixir + OTP that are packaged with OBS/IBS and the one used for development.

Note that this change will need to be propagated through the other repos that use Elixir also.


templates/elixir-ci.yaml
```diff
@@ -14,6 +14,8 @@ env:
   ELIXIR_VERSION: 1.13.4
   OTP_VERSION: 24
   MIX_ENV: test
+  ELIXIR_VERSION_OBS: 1.13.4
```
fabriziosestito, line 17: we could remove the env here and use the matrix directly

I would say we could  use the latest elixir version with the latest OTP, and the obs elixir version + the obs otp version.

jamie-suse: I didn't do that because in the `web` repo's [ci.yaml](https://github.com/trento-project/web/blob/main/.github/workflows/ci.yaml) for example, there is a `test` step and `test-e2e` step for example, and we would have to maintain the versions across both steps.

jamie-suse: (assuming that we are running both versions on both steps, is that correct @fabriziosestito?)

templates/elixir-ci.yaml
```diff
@@ -11,7 +11,7 @@ on:
     types: [start-ci]
 
 env:
-  ELIXIR_VERSION: 1.13.4
+  ELIXIR_VERSION: 1.14.3
   OTP_VERSION: 24
```
fabriziosestito, line 15: latest otp is 25

fabriziosestito (approved):
❤️

PR #19, opened 2023-02-09
Pin version of all CI runners

jamie-suse:
As the title says. Going forward, all  CI runners' versions will be pinned instead of using `-latest` to avoid compatibility issues.


fabriziosestito (approved):

fabriziosestito:
@jamie-suse you need to sign the commit

PR #20, opened 2023-02-09
Add Faker for Wanda

rtorrero:
Initial ADR for Wanda Faker

### DIFF
--- a/adr/0009-add-faker-for-wanda.md
+++ b/adr/0009-add-faker-for-wanda.md
@@ -12,21 +12,18 @@ Since the switch to use Wanda, we currently don't have a way to generate fake ex
 is needed to showcase the functionality of the checks, which is a core aspect of Trento. The current implementation uses live data, which is not ideal
 for demonstration purposes. A Faker-based implementation will provide a safe and controlled environment for demos.
 
-It makes sense to implement this in Wanda rather than in the server for several reasons:
+## Decision
+
+We will implement the faked demo environment data creation in Wanda. These are the main reasons:
 
 - We already have a factory module which takes care of generating such fake data for the purpose of testing.
 - Generating such data in a consistent way is not a trivial task.
 - Generating this data in wanda itself and not in the web brings us closer to simulating the real flow of an execution. We could use mocked agents
   but that would already increase the complexity of a deployment.
 
-## Decision
-
-1.  Create a new Elixir module that implements the ServerBehaviour using the Faker library.
-2.  Add a configuration for the demo environment to use the Faker-based implementation instead of the live implementation.
-3.  Test it by deploy a demo environment using the Faker-based implementation to ensure that it meets the requirements for demonstrating the functionality of the execution server.
-
 ## Consequences
 
-- The demo environment will no longer rely on live data, providing a safer and more controlled environment for demonstrations.
+- The demo environment will have more realistic data without the complexity of executing tests in real infrastructure
+  providing a safer and more controlled environment for demonstrations.
 - The Faker-based implementation may not accurately reflect the behavior of the live implementation, so care must be taken to ensure that the demo environment is representative of the actual system.
 - The Faker-based implementation will require maintenance if the ServerBehaviour changes, to ensure that it continues to accurately reflect the requirements for demonstration purposes.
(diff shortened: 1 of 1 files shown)


adr/0009-add-faker-for-wanda.md
```diff
@@ -0,0 +1,24 @@
...
+
+Date: 2023-02-09
+
+## Status
+
+Accepted
+
+## Context
+
+A demo environment is needed to showcase the functionality of the execution server. The current implementation uses live data, which is not ideal for demonstration purposes. A Faker-based implementation will provide a safe and controlled environment for demos.
```
arbulu89, line 11: I find this `context` paragraph confusing. I would start mentioning that we are talking about wanda (just saying it in the title is not enough in my opinion), and later on, that using real agents to run the evaluation is a problem for demos. 

In fact, the real decision here, and what the ADR should aim for is to tell that we decided to the "faked" code in wanda, and not in the web frontend/backend, or mocking the agents themselves, which we will probably finish doing.

After saying that, I'm struggling to decide if we even want an ADR with this "wanda" faked code now, as we might replace it by mocked agents.

adr/0009-add-faker-for-wanda.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Decision
+
+1.  Create a new Elixir module that implements the ServerBehaviour using the Faker library.
```
arbulu89, line 15: Here the decision shouldn't talk about "code" things. It should talk that we chose to fake the demo evaluations in wanda, instead of other places.

adr/0009-add-faker-for-wanda.md
rtorrero: I've added a bit more context there. I've added some reasons to do thsi in wanda as part of the context. WDYT?

adr/0009-add-faker-for-wanda.md
rtorrero: Since I've discussed this in the context, I think the decision part should document a bit the steps taken to implement this. Do you think it's still needed to add anything else here?

arbulu89: Actually, I would put the bullet list from the `Context` chapter in the `Decision` and remove this. I still don't like having so low level technical details in a ADR.

adr/0009-add-faker-for-wanda.md
```diff
@@ -0,0 +1,32 @@
...
+
+Accepted
+
+## Context
+
+Since the switch to use Wanda, we currently don't have a way to generate fake execution data to be used for our demo environment. The demo environment
+is needed to showcase the functionality of the checks, which is a core aspect of Trento. The current implementation uses live data, which is not ideal
+for demonstration purposes. A Faker-based implementation will provide a safe and controlled environment for demos.
+
+It makes sense to implement this in Wanda rather than in the server for several reasons:
```
arbulu89, line 15: I would put these elements in the decision, starting like:
```
We will implement the faked demo environment data creation in Wanda. These are the main reasons:
# and we list what you wrote here
```

adr/0009-add-faker-for-wanda.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Decision
+
+1.  Create a new Elixir module that implements the ServerBehaviour using the Faker library.
+2.  Add a configuration for the demo environment to use the Faker-based implementation instead of the live implementation.
+3.  Test it by deploy a demo environment using the Faker-based implementation to ensure that it meets the requirements for demonstrating the functionality of the execution server.
+
+## Consequences
+
+- The demo environment will no longer rely on live data, providing a safer and more controlled environment for demonstrations.
```
arbulu89, line 30: ocd: Even though I like this 1st phrase, it is inaccurate, as the current demo env doesn't use live data

arbulu89 (commented):
Hey @rtorrero ,
I think we are going in the right direction.
I vote for putting the 1st bullet points list as the decision, and removing the low level technical details.
Besides that, the text looks good to me

PD: Let's put `Wanda` in capital letters all over the text

rtorrero:
Closing for cleanup, we can reconsider this in the future.

PR #28, opened 2023-05-30
Update rest api versioning strategy

nelsonkopliku:
It looks like we missed to track the policy we decided to follow in regards to API version bump: https://learn.microsoft.com/en-us/linkedin/shared/breaking-change-policy

Updated a pre-existing ADR, not sure we really need a new one.


fabriziosestito (approved):
Great!

PR #35, opened 2023-12-01
Manual installation process documentation

rtorrero:
This PR adds a step-by-step guide to a complete trento deployment without relying on helm/k8s.

### DIFF
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -125,6 +125,11 @@ CREATE USER trento_user WITH PASSWORD 'web-password';
 GRANT ALL PRIVILEGES ON DATABASE trento TO trento_user;
 CREATE DATABASE trento_event_store;
 GRANT ALL PRIVILEGES ON DATABASE trento_event_store TO trento_user;
+
+ALTER DATABASE wanda OWNER TO wanda_user;
+ALTER DATABASE trento OWNER TO trento_user;
+ALTER DATABASE trento_event_store OWNER TO trento_user;
+\l
 \q
 ```
 
@@ -132,8 +137,13 @@ Allow wandadb to be accessible by the docker containers,
 by adding the following line to `/var/lib/pgsql/data/pg_hba.conf`:
 
 ```bash
-host    wanda         postgres          172.17.0.0/16           md5
-host    trento        postgres          172.17.0.0/16           md5
+host    all             all             0.0.0.0/0            md5
+```
+
+Add this line to `/var/lib/pgsql/data/postgresql.conf` to access postgres database
+
+```bash
+listen_addresses = '127.0.0.1, 172.17.0.1'
 ```
 
 Restart postgres to apply the changes:
@@ -142,57 +152,68 @@ Restart postgres to apply the changes:
 systemctl reload postgresql
 ```
 
-### Install rabbitmq
+### Install RabbitMQ
 
 ```bash
 zypper install rabbitmq-server
 ```
 
-Enable and start the rabbitmq service:
+As the agent needs to reach RabbitMQ, allow connections from external hosts.
+Modify `/etc/rabbitmq/rabbitmq.conf` and ensure the following lines are present:
+
+```bash
+listeners.tcp.default = 5672
+```
+
+Add an exception on firewalld:
+
+```bash
+firewall-cmd --zone=public --add-port=5672/tcp --permanent;
+firewall-cmd --reload
+```
+
+As the agent needs to reach RabbitMQ, allow connections from external hosts:
 
 ```bash
 systemctl enable --now rabbitmq-server
 ```
 
-As the agent will need to be able to reach rabbitmq, we need to allow it to connect from external hosts. Modify `/etc/rabbitmq/rabbitmq.conf` and ensure the following lines are present:
+### Configure RabbitMQ
+
+> Note: In order to configurate RabbitMQ for a production system, follow the official suggestions [RabbitMQ guide](https://www.rabbitmq.com/production-checklist.html)
+
+Create a new RabbitMQ user and change the following credentials:
+
+- rabbitmq_user
+- rabbitmq_user_password
+- vhost
 
 ```bash
-listeners.tcp.default = 5672
-...
-management.tcp.ip = 0.0.0.0
+rabbitmqctl add_user trento_user trento_user_password
 ```
 
-In addition to this, we need to allow the port to be accessible from external hosts. Add an exception on firewalld:
+Set role for new user:
 
 ```bash
-firewall-cmd --zone=public --add-port=5432/tcp --permanent
-firewall-cmd --reload
+rabbitmqctl set_user_tags trento_user administrator
 ```
 
-#### Optional: Enable rabbitmq management console
-
-If you want to be able to access rabbitmq management console, the relevant plugin needs to be enabled:
+Verify that the user and the role are correct:
 
 ```bash
-rabbitmq-plugins enable rabbitmq_management
-systemctl restart rabbitmq-server
+rabbitmqctl list_users
 ```
 
-If you are login in from an external host, allow connecting from external hosts (config tweak):
-
-Modify `/etc/rabbitmq/rabbitmq.conf`:
+Create a virtual host:
 
 ```bash
-
-management.tcp.port = 15672
-
+rabbitmqctl add_vhost vhost
 ```
 
-If you are connecting from an external host, add an exception on firewalld:
+Set permissions for the user on the virtual host:
 
 ```bash
-firewall-cmd --zone=public --add-port=15672/tcp --permanent
-firewall-cmd --reload
+rabbitmqctl set_permissions -p vhost trento_user ".*" ".*" ".*"
 ```
 
 ### Install container runtime
@@ -224,7 +245,6 @@ WANDA_SECRET_KEY_BASE=$(openssl rand -out /dev/stdout 48 | base64)
 TRENTO_SECRET_KEY_BASE=$(openssl rand -out /dev/stdout 48 | base64)
 ACCESS_TOKEN_ENC_SECRET=$(openssl rand -out /dev/stdout 48 | base64)
 REFRESH_TOKEN_ENC_SECRET=$(openssl rand -out /dev/stdout 48 | base64)
-
 ```
 
 #### Create a dedicated docker network for trento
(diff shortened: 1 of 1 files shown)


guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+
+This documents covers the step to manually install trento server without relying on the trento helm chart or kubernetes all together. The lastest
+available version of SUSE Linux Enterprise Server for SAP Applications is used as the base operating system, which is 15 SP5 at the time of writing.
+
+## List of dependencies
+
+- prometheus
+- postgresql
+- rabbitmq
+- docker runtime
```
EMaksy, line 11: Maybe add an installation step for installing docker?
Just in case to ensure that the reader has everything

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Installation
+
+### Install prometheus (Optional)
+
+#### Option 1: Use existing installation
+
+If you already have a prometheus server running, you can skip this step. , you can skip the inst
+
```
EMaksy, line 20: How should the reader install Prometheus ? Maybe adding some commands or a reference to official docs?

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+#### Option 2: Do not use prometheus
+
+If you don't have a Prometheus instance running and you don't want to install one, you can continue without doing so. The downside is that trento will
+not be able to collect metrics from prometheus and the integrated graphs will not work.
```
EMaksy, line 24: The explanation would be more helpful if we move it in line 16 under the ### Install prometheus (Optional)" header. 
As a reader i would know why i would need Prometheus and then with decide with which option to use or not use.

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+
+Allow prometheus to be accessible from docker. Add an exception on firewalld:
+
+```bash
+firewall-cmd --zone=public --add-port=9090/tcp --permanent
+firewall-cmd --reload
+```
+
+### Install grafana (Optional)
+
```
EMaksy, line 72: A small explanation why a reader would need to have grafana ?

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+zypper in postgresql-server
+```
+
+Enable and start the postgresql service:
+
+```bash
+systemctl enable --now postgresql
+```
+
+Initialize the postgresql databases (be sure to update the passwords accordingly):
```
EMaksy, line 112: Maybe instead of (...) something like a note  or hint?

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+CREATE DATABASE trento;
+CREATE USER trento_user WITH PASSWORD 'web-password';
+GRANT ALL PRIVILEGES ON DATABASE trento TO trento_user;
+CREATE DATABASE trento_event_store;
+GRANT ALL PRIVILEGES ON DATABASE trento_event_store TO trento_user;
+\q
+```
+
+Allow wandadb to be accessible by the docker containers,
+by adding the following line to `pg_hba.conf`:
```
EMaksy, line 129: Maybe add a path for the user to be able to directly edit it ? without researching? 
on my side it was in `/var/lib/pgsql/data/pg_hba.conf`

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+
+```bash
+systemctl enable --now rabbitmq-server
+```
+
+As the agent will need to be able to reach rabbitmq, we need to allow it to connect from external hosts. Modify `/etc/rabbitmq/rabbitmq.conf` and ensure the following lines are present:
+
+```bash
+listeners.tcp.default = 5672
+...
```
EMaksy, line 158: do we need the ...  ?

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+This documents covers the step to manually install trento server without relying on the trento helm chart or kubernetes all together. The lastest
+available version of SUSE Linux Enterprise Server for SAP Applications is used as the base operating system, which is 15 SP5 at the time of writing.
+
+## List of dependencies
+
+- prometheus
+- postgresql
+- rabbitmq
+- docker runtime
+
```
EMaksy, line 12: What about grafana?

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
+## Scope
+
+This documents covers the step to manually install trento server without relying on the trento helm chart or kubernetes all together. The lastest
```
EMaksy, line 3: Should we also mention the ansible option here? 
https://github.com/trento-project/ansible

guides/manual-installation.md
```diff
@@ -0,0 +1,279 @@
...
+    -e ACCESS_TOKEN_ENC_SECRET=$ACCESS_TOKEN_ENC_SECRET \
+    -e REFRESH_TOKEN_ENC_SECRET=$REFRESH_TOKEN_ENC_SECRET \
+    -e ADMIN_USERNAME='admin' \
+    -e ADMIN_PASSWORD='test1234' \
+    -e GRAFANA_PASSWORD='trento' \
+    -e ENABLE_API_KEY='true' \
+    --entrypoint /bin/sh \
+    registry.suse.com/trento/trento-web:2.2.0 \
+    -c "/app/bin/trento eval 'Trento.Release.init()' && /app/bin/trento start"
+```
```
EMaksy, line 279: just an idea, add some commands to ensure that the containers are running fine. 
Also how to actually access trento web would be great!

guides/manual-installation.md
```diff
@@ -0,0 +1,326 @@
...
+- rabbitmq
+- docker runtime
+
+## Installation
+
+### Install prometheus (Optional)
+
+#### Option 1: Use existing installation
+
+If you already have a prometheus server running, you can skip this step. , you can skip the inst
```
EMaksy, line 20: Remove duplication?

guides/manual-installation.md
```diff
@@ -0,0 +1,326 @@
...
+```bash
+listeners.tcp.default = 5672
+...
+management.tcp.ip = 0.0.0.0
+```
+
+In addition to this, we need to allow the port to be accessible from external hosts. Add an exception on firewalld:
+
+```bash
+firewall-cmd --zone=public --add-port=5432/tcp --permanent
```
rtorrero, line 168: typo, wrong port

guides/manual-installation.md
rtorrero: I added a link to the official prometheus site, in case the user doesn't want to follow our suggestions to install it, he can go there and follow instructions there.

guides/manual-installation.md
```diff
@@ -0,0 +1,505 @@
...
+Enable and start the prometheus service:
+
+```bash
+systemctl enable --now prometheus
+```
+
+Allow prometheus to be accessible from docker. Add an exception on firewalld:
+
+```bash
+firewall-cmd --zone=public --add-port=9090/tcp --permanent
```
CDimonaco, line 74: With the zone public, we are sure that they can't access prometheus from outside the machine? We need just to make sure that is accessible from the docker subnet

guides/manual-installation.md
```diff
@@ -0,0 +1,505 @@
...
+```sql
+CREATE DATABASE wanda;
+CREATE DATABASE trento;
+CREATE DATABASE trento_event_store;
+```
+
+**Step 3:** Create the users:
+
+```sql
+CREATE USER wanda_user WITH PASSWORD 'wanda-password';
```
CDimonaco, line 110: Maybe it's better with encrypted password, but idk if it's a default in the new versions of postgres

guides/manual-installation.md
```diff
@@ -0,0 +1,505 @@
...
+
+Enable and start postgresql:
+
+```bash
+systemctl enable --now postgresql
+```
+
+#### Configure postgresql
+
+**Step 1:** Start `psql` to open a connection to the database:
```
CDimonaco, line 92: "with a user with the correct privileges to run the commands", idk how to phrase it

guides/manual-installation.md
```diff
@@ -0,0 +1,505 @@
...
+GRANT ALL ON SCHEMA public TO trento_user;
+\q
+```
+
+**Step 5:** Allow the docker containers to connect to their respective databases by adding the following in `/var/lib/pgsql/data/pg_hba.conf`:
+
+```bash
+host    wanda           wanda_user      0.0.0.0/0               md5
+host    trento          trento_user     0.0.0.0/0               md5
+host    trento_event_store      trento_user     0.0.0.0/0       md5
```
CDimonaco, line 131: I think we should suggest also here by default to use the docker subnet

guides/manual-installation.md
```diff
@@ -0,0 +1,505 @@
...
+
+Enable and start Docker:
+
+```bash
+systemctl enable --now docker
+```
+
+### Deploy trento-wanda and trento-web components on docker:
+
+#### Create the secrets
```
CDimonaco, line 234: Please don't refer as them by "secrets" this could be misleading in a docker context, just env variables

guides/manual-installation.md
```diff
@@ -0,0 +1,505 @@
...
+REFRESH_TOKEN_ENC_SECRET=$(openssl rand -out /dev/stdout 48 | base64)
+```
+
+#### Create a dedicated docker network for trento
+
+```bash
+docker network create trento-net
+```
+
+#### Install trento-wanda on docker:
```
CDimonaco, line 249: Before giving the shell command for containers, we should say that they should keep all the env variables secure as they are sensible secrets, and they should follow all the security best practices and if they want more security to use a docker secrets approach.

At the end is just an example

CDimonaco (changes requested):
Just some comments, but generally speaking I think it's a good job!

guides/manual-installation.md
rtorrero: True, it should be `--zone=docker`. I'll change it :+1:

guides/manual-installation.md
rtorrero: According to https://www.postgresql.org/docs/10/release-10.html, unencrypted passwords where deprecated in postgresql 10:
>Remove the ability to store unencrypted passwords on the server (Heikki Linnakangas)

>The [password_encryption](https://www.postgresql.org/docs/10/runtime-config-connection.html#GUC-PASSWORD-ENCRYPTION) server parameter no longer supports off or plain. The UNENCRYPTED option is no longer supported in CREATE/ALTER USER ... PASSWORD. Similarly, the --unencrypted option has been removed from createuser. Unencrypted passwords migrated from older versions will be stored encrypted in this release. The default setting for password_encryption is still md5.

CDimonaco (approved):

PR #38, opened 2024-01-08
Enrich manual install guide

EMaksy:
Just small improvements (WIP)

### DIFF
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -1,3 +1,5 @@
+# Manual Installation of Trento
+
 ## Scope
 
 This documents covers the step to manually install trento server without relying on the trento helm chart or kubernetes all together. The lastest
@@ -5,15 +7,15 @@ available version of SUSE Linux Enterprise Server for SAP Applications is used a
 
 For other installation options, check:
 
-- https://github.com/trento-project/helm-charts/
-- https://github.com/trento-project/ansible (WIP)
+- [Trento Ansible](https://github.com/trento-project/ansible) (WIP)
+- [Trento Helm-Charts](https://github.com/trento-project/helm-charts/)
 
 ## List of dependencies
 
-- prometheus (optional)
-- postgresql
-- rabbitmq
-- docker runtime
+- [Prometheus](https://prometheus.io/) (optional)
+- [PostgreSQL](https://www.postgresql.org/)
+- [RabbitMQ](https://rabbitmq.com/)
+- [Docker](https://www.docker.com/)
 
 ## Installation
 
@@ -77,6 +79,8 @@ firewall-cmd --reload
 
 ### Install postgresql
 
+> Note: This guide was tested with the PostgreSQL server package version **_16-150500.10.3.2_** from the repository SLE-Module-Packagehub-Subpackages15-SP5-Updates. Using a different version of the postgresql-server package may require different steps or configurations, specially when changing the major number. Refer to the official [postgress documentation](https://www.postgresql.org/docs/) for further guidance
+
 ```bash
 zypper in postgresql-server
 ```
@@ -107,8 +111,8 @@ CREATE DATABASE trento_event_store;
 **Step 3:** Create the users:
 
 ```sql
-CREATE USER wanda_user WITH PASSWORD 'wanda-password';
-CREATE USER trento_user WITH PASSWORD 'web-password';
+CREATE USER wanda_user WITH PASSWORD 'wanda_password';
+CREATE USER trento_user WITH PASSWORD 'web_password';
 ```
 
 **Step 4:** Grant required privileges to the users and close the connection:
@@ -126,9 +130,9 @@ GRANT ALL ON SCHEMA public TO trento_user;
 **Step 5:** Allow the docker containers to connect to their respective databases by adding the following in `/var/lib/pgsql/data/pg_hba.conf`:
 
 ```bash
-host    wanda           wanda_user      172.17.0.0/16               md5
-host    trento          trento_user     172.17.0.0/16               md5
-host    trento_event_store      trento_user     172.17.0.0/16       md5
+host   wanda                wanda_user    172.17.0.0/16   md5
+host   trento               trento_user   172.17.0.0/16   md5
+host   trento_event_store   trento_user   172.17.0.0/16   md5
 ```
 
 **Step 6:** Allow PostgreSQL to bind on all interfaces `/var/lib/pgsql/data/postgresql.conf` by changing the following line:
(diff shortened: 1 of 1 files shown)


guides/manual-installation.md
```diff
@@ -1,19 +1,21 @@
+# Manual Installation of Trento
```
arbulu89, line 1: Already late, as the file is named as that, but for `Manual` word wouldn't be needed.
This would be `Installing Trento` guide.

guides/manual-installation.md
```diff
@@ -1,19 +1,21 @@
...
 
 This documents covers the step to manually install trento server without relying on the trento helm chart or kubernetes all together. The lastest
 available version of SUSE Linux Enterprise Server for SAP Applications is used as the base operating system, which is 15 SP5 at the time of writing.
 
 For other installation options, check:
 
-- https://github.com/trento-project/helm-charts/
-- https://github.com/trento-project/ansible (WIP)
+- [Trento Ansible](https://github.com/trento-project/ansible) (WIP)
+- [Trento Helm-Charts](https://github.com/trento-project/helm-charts/)
```
arbulu89, line 11: `Trento Helm Charts`

guides/manual-installation.md
```diff
@@ -1,19 +1,21 @@
...
 
 ## List of dependencies
 
-- prometheus (optional)
-- postgresql
-- rabbitmq
-- docker runtime
+- [Prometheus](https://prometheus.io/) (optional)
```
arbulu89, line 15: Let's put prometheus last, as it is optional

guides/manual-installation.md
```diff
@@ -77,6 +79,8 @@ firewall-cmd --reload
 
 ### Install postgresql
 
+> Note: This guide was tested with the PostgreSQL server package version **_16-150500.10.3.2_** from the repository SLE-Module-Packagehub-Subpackages15-SP5-Updates. Using a different version of the postgresql-server package may require different steps or configurations, specially when changing the major number. Refer to the official [postgress documentation](https://www.postgresql.org/docs/) for further guidance
```
arbulu89, line 82: We shouldn't promote a package version coming from package hub.
PostgreSQL is available in the base module `sle-module-basesystem`. We should check this.
And postresql-server is available in `sle-module-server-applications`

guides/manual-installation.md
```diff
@@ -240,9 +242,11 @@ REFRESH_TOKEN_ENC_SECRET=$(openssl rand -out /dev/stdout 48 | base64)
 docker network create trento-net
 ```
 
-#### Install trento-wanda on docker:
+#### Install trento on docker
```
arbulu89, line 245: `trento` -> `Trento`

arbulu89 (commented):
Some comments. The content looks correct otherwise

guides/manual-installation.md
```diff
@@ -77,6 +78,9 @@ firewall-cmd --reload
 
 ### Install postgresql
 
+> Note: This guide was tested with the PostgreSQL version **15.5**
```
arbulu89, line 81: This 15.5 version looks suspicious. We are already in `16`.
@rtorrero could you confirm this?

arbulu89 (approved):

guides/manual-installation.md
rtorrero: I just checked in the ansible-installed machine with sles 15 sp5 and I get:
```
localhost:~ # psql --version
psql (PostgreSQL) 15.5
```

I see `postgresql16-server` indeed available to be installed in zypper, but when using the generic package name `postgresql-server` it still pulls `postgresql15-server` by default.

rtorrero (approved):
LGTM @EMaksy, thanks!

guides/manual-installation.md
EMaksy: I took a quick look at the spec file and per default on sle15 sp5 we get postgresql15-server installed :)

arbulu89: @EMaksy But the installed package version on zypper is still `16`. Maybe we need to put a comment about that, or ask to the package maintainer why is this happening.

guides/manual-installation.md
```diff
@@ -1,19 +1,20 @@
+# Installation of Trento
+
 ## Scope
 
-This documents covers the step to manually install trento server without relying on the trento helm chart or kubernetes all together. The lastest
-available version of SUSE Linux Enterprise Server for SAP Applications is used as the base operating system, which is 15 SP5 at the time of writing.
+This document covers the steps to manually install trento server without relying on the trento helm chart or kubernetes all together. The latest available version of SUSE Linux Enterprise Server for SAP Applications is used as the base operating system, which is 15 SP5 at the time of writing.
```
arbulu89, line 5: I would remove any mention to the the helm chart. Having that way first is just a coincidence.
Just `This document covers the steps to manually install Trento server` would be enough in my opinion

arbulu89 (approved):

guides/manual-installation.md
EMaksy: I will ping the maintainer :+1:

EMaksy: Just a quick update about the Postgres version: 

The postgresql package is a wrapper package which pulls in the main package and postgresqlXX usually.

The version of the package is unrelated to the version of Postgres which is used.

For SLES15 SP5 it is Postgres version 15.
For SLE15 SP6 it will be Postgres version 16.

So I think we are good to go with 15.5 :+1:

CDimonaco (approved):

guides/manual-installation.md
```diff
@@ -39,29 +45,31 @@ Enable PackageHub repository:
...
 
-After a successful registration, you can install prometheus using zypper:
+Install prometheus using zypper:
 
 ```bash
 zypper in golang-github-prometheus-prometheus
 ```
 
-We will be prompted that a missing dependency can't be satisfied:
+Missing dependency can't be satisfied **(ONLY FOR SLE15 SP5)**:
```
rtorrero, line 63: I noticed while adding support for SP4 in the Ansible playbooks that SP4 also requires the Prometheus workaround. Could you change this to also include SLES15 SP4?

rtorrero (changes requested):
Just that small change, everything else LGTM!

guides/manual-installation.md
rtorrero: Another minor thing, could you add this commit or something similar?: https://github.com/trento-project/docs/commit/6950503017540f6eaccc94870008df2d56212ba1 see task TRNT-2689

rtorrero (approved):
LGTM @EMaksy thanks!

PR #39, opened 2024-01-22
Add prometheus configuration

EMaksy:
This pr will enrich the existing manual installation guide with missing Prometheus configuration.

Thanks to @arbulu89 for spotting this!

### DIFF
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -518,6 +518,7 @@ server {
         # The Important Websocket Bits!
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
+        proxy_set_header Origin "";
 
         proxy_pass http://localhost:4000;
     }
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -90,6 +90,8 @@ scrape_configs:
         url: http://localhost:4000/api/prometheus/targets
 ```
 
+> Note: **localhost:4000** in **url: http://localhost:4000/api/prometheus/targets** refers to the location where Trento web docker container is running.
+
 Enable and start the prometheus service:
 
 ```bash
@@ -317,7 +319,7 @@ docker run -d \
  -e DATABASE_URL=ecto://trento_user:web_password@host.docker.internal/trento \
  -e EVENTSTORE_URL=ecto://trento_user:web_password@host.docker.internal/trento_event_store \
  -e ENABLE_ALERTING=false \
- -e PROMETHEUS_URL='localhost:9090' \
+ -e PROMETHEUS_URL='http://host.docker.internal:9090' \
  -e SECRET_KEY_BASE=$TRENTO_SECRET_KEY_BASE \
  -e ACCESS_TOKEN_ENC_SECRET=$ACCESS_TOKEN_ENC_SECRET \
  -e REFRESH_TOKEN_ENC_SECRET=$REFRESH_TOKEN_ENC_SECRET \
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -31,7 +31,7 @@ Other installation options:
 
 #### <a id="prometheus_install_option_1"></a>Option 1: Use existing installation
 
-If you are planning to [install Prometheus](https://prometheus.io/docs/prometheus/latest/installation/) or have an existing Prometheus server, ensure to set the PROMETHEUS_URL environment variable with your Prometheus server's URL as part of the Docker command when creating the trento-web container.
+If you have an [existing Prometheus server](https://prometheus.io/docs/prometheus/latest/installation/), ensure to set the PROMETHEUS_URL environment variable with your Prometheus server's URL as part of the Docker command when creating the trento-web container.
 
 > Note: Minimal required prometheus version is **2.28.0**
 
@@ -51,7 +51,7 @@ SUSEConnect --product PackageHub/15.5/x86_64
 > Note: SLE15 SP3 requires a provided prometheus server. The version available through **SUSEConnect --product PackageHub/15.3/x86_64** is outdated and is not compatible with Trento's prometheus configuration.
 > Refer to [Option 1: Use existing installation option](#prometheus_install_option_1) for SLE 15 SP3.
 
-> Note: Using a different Service Pack than SP5 requires to change repository: SLE15 SP4: `SUSEConnect --product PackageHub/15.4/x86_64`
+> Note: Using SLE15 SP4 requires changing the repository `SUSEConnect --product PackageHub/15.4/x86_64`
 
 Add the prometheus user/group:
 
@@ -104,7 +104,7 @@ Enable and start the prometheus service:
 systemctl enable --now prometheus
 ```
 
-Allow prometheus to be accessible from docker and add an exception on firewalld **(ONLY if firewalld is enabled)**:
+Allow prometheus to be accessible from docker and add an exception on firewalld **(ONLY if firewalld is running)**:
 
 ```bash
 firewall-cmd --zone=docker --add-port=9090/tcp --permanent
@@ -195,7 +195,7 @@ Modify `/etc/rabbitmq/rabbitmq.conf` and ensure the following lines are present:
 listeners.tcp.default = 5672
 ```
 
-Add an exception on firewalld **(ONLY if firewalld is enabled)**:
+Add an exception on firewalld **(ONLY if firewalld is running)**:
 
 ```bash
 firewall-cmd --zone=public --add-port=5672/tcp --permanent;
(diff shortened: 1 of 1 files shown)


guides/manual-installation.md
```diff
@@ -70,6 +70,26 @@ Problem: nothing provides 'group(prometheus)' needed by the to be installed gola
...
+  - job_name: "http_sd_hosts"
+    honor_timestamps: true
+    scrape_interval: 30s
+    scrape_timeout: 30s
+    scheme: http
+    follow_redirects: true
+    http_sd_configs:
+      - follow_redirects: true
+        refresh_interval: 1m
+        url: http://localhost:4000/api/prometheus/targets
```
arbulu89, line 90: We should point out that this `localhost:4000` is where the web container is running.

guides/manual-installation.md
```diff
@@ -297,7 +317,7 @@ docker run -d \
  -e DATABASE_URL=ecto://trento_user:web_password@host.docker.internal/trento \
  -e EVENTSTORE_URL=ecto://trento_user:web_password@host.docker.internal/trento_event_store \
  -e ENABLE_ALERTING=false \
- -e PROMETHEUS_URL='http://host.docker.internal:9090' \
+ -e PROMETHEUS_URL='localhost:9090' \
```
arbulu89, line 320: why this change?
this `PROMETHEUS_URL` is used by the trento web container, which as it is running in a conatiner, having `host.docker.internal` says that prometheus is running in the host machine.
With `localhost`, you are saying that prometheus is running in the same container.

guides/manual-installation.md
```diff
@@ -419,7 +439,7 @@ certbot --nginx -d example.com -d www.example.com
 zypper install nginx
 ```
 
-**Step 2**: Add firewall exceptions for HTTP and HTTPS:
+**Step 2**: Add firewalld exceptions for HTTP and HTTPS **(ONLY if firewalld is enabled)**:
```
arbulu89, line 442: @rtorrero has commented that to have docker engine running, firewalld must be running.
I don't know how accurate is that, but let's have an eye

arbulu89 (changes requested):

guides/manual-installation.md
EMaksy: My bad, thanks! I understood it before exact in an opposite way!

guides/manual-installation.md
EMaksy: Docker engine not work without firewalld ? This is new to me :+1: 
@rtorrero maybe you can provide more info on that.

CDimonaco: It's some misconfiguration issue, docker itself does not have any dependencies on the firewall, it's some quirky iptables stuff I guess :/

abravosuse: I installed Trento Server in a VM where firewalld was not running.

abravosuse (approved):

guides/manual-installation.md
rtorrero: The issue only appears when firewalld is uninstalled and not rebooted. I think it's safe to assume this is an edge case that can be ignored.

guides/manual-installation.md
```diff
@@ -70,6 +70,28 @@ Problem: nothing provides 'group(prometheus)' needed by the to be installed gola
...
+  evaluation_interval: 10s
+
+scrape_configs:
+  - job_name: "http_sd_hosts"
+    honor_timestamps: true
+    scrape_interval: 30s
+    scrape_timeout: 30s
+    scheme: http
+    follow_redirects: true
+    http_sd_configs:
```
rtorrero, line 87: This `http_sd_config` (and I think the `follow_redirects` too) is apparently not supported in the prometheus version we can get in SP3

rtorrero (commented):
@EMaksy Good stuff! I think it just misses the details about the scraper config not currently working on SP3 (haven't checked in SP4 but it's probably good there, I will check today). 

Basically, this means we either address this by submitting an updated prometheus package to the sp3 stream or we add a temporary note that prometheus is currently not supported in this scenario. @arbulu89 @abravosuse  wdyt?

guides/manual-installation.md
arbulu89: @rtorrero Let's keep thas as is while we submit backport prometheus until SP3

EMaksy:
SP4 works fine :)

About SP3, sadly the packaged version is too old. 
**SUSEConnect --product PackageHub/15.3/x86_64** repository provides prometheus version 2.18.0. 
Minimal requirement of Prometheus for our configuration would be 2.28.0.

As Prometheus is optional, I would reference the reader to use an existing Prometheus server and not rely on PackageHub
What do you think?

guides/manual-installation.md
```diff
@@ -45,9 +49,12 @@ Enable PackageHub repository:
 SUSEConnect --product PackageHub/15.5/x86_64
 ```
 
-> Note: Using a different Service Pack then SP5 requires to change repository: [SLE15 SP3: `SUSEConnect --product PackageHub/15.3/x86_64`, SLE15 SP4: `SUSEConnect --product PackageHub/15.4/x86_64`]
+> Note: SLE15 SP3 requires a provided prometheus server. The version available through **SUSEConnect --product PackageHub/15.3/x86_64** is outdated and is not compatible with Trento's prometheus configuration.
+> Refer to [Option 1: Use existing installation option](#prometheus_install_option_1) for SLE 15 SP3.
+
+> Note: Using a different Service Pack then SP5 requires to change repository: SLE15 SP4: `SUSEConnect --product PackageHub/15.4/x86_64`
```
arbulu89, line 55: typo, `then` -> `than`

guides/manual-installation.md
```diff
@@ -45,9 +49,12 @@ Enable PackageHub repository:
 SUSEConnect --product PackageHub/15.5/x86_64
 ```
 
-> Note: Using a different Service Pack then SP5 requires to change repository: [SLE15 SP3: `SUSEConnect --product PackageHub/15.3/x86_64`, SLE15 SP4: `SUSEConnect --product PackageHub/15.4/x86_64`]
+> Note: SLE15 SP3 requires a provided prometheus server. The version available through **SUSEConnect --product PackageHub/15.3/x86_64** is outdated and is not compatible with Trento's prometheus configuration.
```
arbulu89, line 52: I don't understand the `SLE15 SP3 requires a provided prometheus server`. Does this mean that you need to have your own prometheus installation?
`SLE15 SP3 requires an already installed prometheus server, as the given prometheus version for this service pack is outdated`, or something similar.
Anyway, i know that that's true, but it sound terrible in a guide :sweat_smile:

arbulu89 (approved):
It looks good.
I think we could maybe insist on shipping the required prometheus version in SP3 package hub somehow, maybe scalating a bit, because if i were the user and read this, I wouldn't be really convinced...

abravosuse (changes requested):
Hey @EMaksy,

I tried the manual installation procedure on SLES4SAP 15 SP5 (Azure image), with Prometheus and self-signed certificate.

First a couple of minor things on wording:

- “If you are planning to [install Prometheus](https://prometheus.io/docs/prometheus/latest/installation/) or have an existing Prometheus server”
This sentence is in section “Option 1: Use existing installation”. Therefore I find the first part of it “If you are planning to [install Prometheus](https://prometheus.io/docs/prometheus/latest/installation/)” a little bit contradictory.

- “Note: Using a different Service Pack than SP5 requires to change repository: SLE15 SP4:
Since this note is specific to SP4 I would rephrase it to reflect so. Something like “If you are using SP4 change the repository accordingly”

- “ONLY if firewalld is enabled” 
I would replace this with ‘ONLY if firewalld is running”. You might have firewalld enabled but stopped for some reason.

Now some more important stuff that I noticed in my setup:

- The console is not updated automatically when the agent status changes. For example, if I stop the agent, I need to refresh the console in order to see that the agent has no heartbeat. The same when I start it back up.
- The CPU and the Memory graphs in the host details view show a "localhost refused to connect" message. Should I replace "localhost" in the `prometheus.yml` file with something else? 

image

Thank  you!

abravosuse:
Something I forgot to mention yesterday @EMaksy : when deploying wanda container we use `registry.suse.com/trento/trento-wanda:1.2.0' but when deploying web container we use `registry.suse.com/trento/trento-web:latest`. Why a specific tag in one case  and `latest` in the other?

EMaksy:
Hey @abravosuse, thanks for testing. I improved wording, thanks for the suggestions!

I changed the container version to a fixed version.

Sadly no graphs without grafana, that's the reason. 
We [removed grafana](https://github.com/trento-project/docs/commit/00cbf4bf9a735a32d978eef9aaab2f96156adf9a) from the manual guide, as the upcoming release will introduce new charts without Grafana. (Also Grafana package is not supported)

About that the console is not updated automatically when the agent status changes, well we have an open bug for it.
There is a problem that on a HTTPS connection the web socket does not work.

arbulu89:
> About that the console is not updated automatically when the agent status changes, well we have an open bug for it.
There is a problem that on a HTTPS connection the web socket does not work.

@CDimonaco in short, do you know the status of this? As far as I know we already had a bug ticket, and some workaround applied in the ansible playbook

CDimonaco:
> > About that the console is not updated automatically when the agent status changes, well we have an open bug for it.
> There is a problem that on a HTTPS connection the web socket does not work.
> 
> @CDimonaco in short, do you know the status of this? As far as I know we already had a bug ticket, and some workaround applied in the ansible playbook

The bug ticket is to do, but we have a workaround in the ansible playbook. It's one line in the nginx configuration, we should add the same line here in the docs

EMaksy:
Hey @abravosuse, the nginx config is updated now with @CDimonaco proposed workaround. Web sockets are working again!
image

Feel free to test :+1:

PR #42, opened 2024-03-14
Add RPM installation method chapter

arbulu89:
The new RPM installation method chapter.

### DIFF
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -18,9 +18,9 @@ Other installation options:
 
 ## List of dependencies
 
-- [Docker](https://www.docker.com/)
 - [PostgreSQL](https://www.postgresql.org/)
 - [RabbitMQ](https://rabbitmq.com/)
+- [Docker](https://www.docker.com/) (optional)
 - [Prometheus](https://prometheus.io/) (optional)
 
 ## Installation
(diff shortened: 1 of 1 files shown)


guides/manual-installation.md
```diff
@@ -234,7 +234,101 @@ Set permissions for the user on the virtual host:
...
+
+Enable and start the services:
+
+```bash
+systemctl enable --now trento-web trento-wanda
+```
+
+#### Monitor the services
+
+In order to check if the services are up and running properly, `journalctl` can be used.
```
EMaksy, line 321: What do you think about making it a bit shorter?

Check if the services are up and running properly by using `journalctl`

guides/manual-installation.md
```diff
@@ -234,7 +234,101 @@ Set permissions for the user on the virtual host:
 rabbitmqctl set_permissions -p vhost trento_user ".*" ".*" ".*"
 ```
 
-### Install Docker container runtime
+### Install Trento server components
+
+Trento server components are available in 2 different installation formats: RPM packages and Docker images.
+Each of them have a different installation process, but the end result is the same. To choose between any of them
```
EMaksy, line 240: I think it should be changed like this:  Each of them `has` ...

EMaksy (approved):
Hey man good!
I left some minor suggestions but beside this LGTM!!!

abravosuse:
@arbulu89 here you have some comments/remarks regarding the installation procedures using rpms. For context, I have tried it out in three Azure VMs, one with SLES for SAP 15 SP3, another one with SLES for SAP 15 SP4 and a last one with SLES for SAP 15 SP5. In the case of SP3, I have not installed Prometheus:

- [SP4, SP5] Section Install Prometheus Option 2. The note "Note: localhost:4000 in url: http://localhost:4000/api/prometheus/targets refers to the location where Trento web docker container is running" is specific to the docker installation. If the note is also valid for the rpm installation, we should say so. If not, we should also clarify it.
- [All] Postgresql installation: the version that gets installed by default is 13.14 / 14.11 / 15.6 which differs from the one specified in the guide (13.9 / 14.10 / 15.6). The warning about using a different version of postgres might put the user in a dilemma.: whether to continue with these versions or uninstall them and install the ones that have been tested.
- [All] Configure Postgres Step 1. If you switch to user postgres with `su postgres`, then you get a permission error when executing `psql`. You need to switch with `su - postgres`.
- [All] Configure Postgres Step 5. In the case of an rpm installation we need to install the following lines on top of all the host entries:
`host   wanda                wanda_user    0.0.0.0/0   md5
host   trento               trento_user   0.0.0.0/0   md5
host   trento_event_store   trento_user   0.0.0.0/0   md5`

- [All] Configure Postgres Step 7. The previous steps have been executed as user postgres but this one must be executed as root. It may be worth mentioning that we are back to  use user root.
- [All] trento web confiiguration file. Parameter PORT should be set to 4000 not 400.
- [All] Accessing the Trento Web UI. I would put this section before section Testing the setup, since you need the API key to configure the agents.

abravosuse:
Thank you Xabi. Let's leave the part about `postgresql` versions AS-IS then.

PR #43, opened 2024-04-04
Add entry explaining the sap system and database split

arbulu89:
New ADR recording the decisions we took to handle the multi-tenant epic.
It basically explains the decision to split the SAP system and database aggregates.
Most the decisions we took are based in exploratory work and commanded docs.

Handling old SAP system events has been the most challenging thing, that's why many of the comments are related to this.
Everything else is pretty straight forward.

The handling of old events is based on this: https://github.com/trento-project/web/pull/2485


adr/0011-sap-system-database-aggregate-split.md
```diff
@@ -0,0 +1,33 @@
...
+
+In this scenario, when a multi-tenant database is discovered, it only can manage a unique SAP system and its application instances, making the multi-tenant scenario not supported.
+
+This must be reverted, as a multi-tenant scenario is pretty common, and many deployments have them.
+
+## Decision
+
+The currently existing `sap_system.ex` aggregate will be separated in 2, creating a new `database.ex` aggregate that will handle databases individually. This way, the lifecycle of a database is extracted from the SAP system one, making them independent. Having this, a new 1 to N relationship will be created between the database and SAP system aggregates (read models being more precise). Now on, the SAP system ID will be calculated combining the database ID and its tenant name, creating a new and unique ID for each SAP system.
```
janvhs, line 19: Changed the style a bit and fixed two grammatical issues.
See if you prefer it.

```suggestion
The currently existing `sap_system.ex` aggregate will be split into seperate modules, creating a new `database.ex` aggregate that will handle databases individually. This way, the lifecycle of a database is extracted from the lifecycle of the SAP system, making them independent. Having this, a new 1 to N relationship will be created between the database and SAP system aggregates. Now the read models are more precise. From now on, the SAP system ID will be calculated by combining the database ID and its tenant name, creating a new and unique ID for each SAP system.
```

adr/0011-sap-system-database-aggregate-split.md
```diff
@@ -0,0 +1,33 @@
...
+
+The relationship between database and SAP systems will continue existing in many senses, and most of the related events will be handled by event handlers. For example, when a database is deregistered, or its health updated, these events will be listened, and once they happen, they will affect the SAP system aggregate via commands. At the end, both aggregates continue being pure, and connected through this kind of mechanisms.
```
janvhs, line 21: > and most of the related events will be handled by event handlers.

Does this mean the messages will still be handled by the same event handlers, effectively sharing some handlers?

janvhs, line 21:
```suggestion
The relationship between database and SAP systems will continue existing in many senses, and most of the related events will be handled by event handlers. For example, when a database is deregistered, or its health updated, these events will be handled, and once they happen, they will affect the SAP system aggregate via commands. At the end, both aggregates continue being pure and connected through this kind of mechanism.
```

adr/0011-sap-system-database-aggregate-split.md
```diff
@@ -0,0 +1,33 @@
...
+
+In order to deal with all of these old events, besides upcasting, we have considered a couple of options when old events and IDs are mixed up. Initially, [prefixing](https://hexdocs.pm/commanded/Commanded.Commands.Router.html#module-define-aggregate-identity) the `sap_system.ex` aggregate routing it with a `sap-system-` prefix looked a good option. It removes the issues you could have when SAP system commands with old IDs now belonging to databases are dispatched, as they are routed with the prefix together with the ID. We have discarded this option, as it converts the SAP system stream different from the others, and hides a bit what's going on in these scenarios. Instead of this solution, as the unique real scenario when this can happen is when the `DeregisterApplicationInstance` is dispatched by the deregistration process manager, we will just handle this particular command. The command will be enriched in a middleware protocol, so commands with IDs now belonging to databases will be ignored.
```
janvhs, line 25: I'd reword and split this sentence into two, because it helps with readability of a complex scenario.

```suggestion
In order to deal with all of these old events, besides upcasting, we have considered a couple of options when old events and IDs are mixed up. Initially, [prefixing](https://hexdocs.pm/commanded/Commanded.Commands.Router.html#module-define-aggregate-identity) the `sap_system.ex` aggregate and routing it with a `sap-system-` prefix looked a good option. It removes issues that could arise when SAP system commands with old IDs, now belonging to databases, are dispatched, as they are routed with the prefix together with the ID. We have discarded this option, as it converts the SAP system stream different from the others, and hides a bit what's going on in these scenarios. We decided against this solution, as it realistically only occours when the `DeregisterApplicationInstance` is dispatched by the deregistration process manager. Therefore, we will just handle this particular command. The command will be enriched in a middleware protocol, so commands with IDs, now belonging to databases, will be ignored.
```

janvhs (commented):
Your explanation is great in providing context and describing the agreed on solution! I feel like I clearly understand the change and why it is necessary.
However, I suggest rewording some sentences and fixing the grammatical mistakes I highlighted.

adr/0011-sap-system-database-aggregate-split.md
```diff
@@ -0,0 +1,33 @@
...
+
+Accepted
+
+## Context
+
+In the initial implementation of Trento, the SAP system and database states are stored in the same `sap_system.ex` aggregate. This is making the assumption that the relationship between a SAP system and a database is of 1 to 1, meaning that a database is only used by one SAP system. This is not an accurate description of the reality, as a database can handle multiple SAP systems in a multi-tenant environment.
+
+In this scenario, when a multi-tenant database is discovered, it only can manage a unique SAP system and its application instances, making the multi-tenant scenario not supported.
+
+This must be reverted, as a multi-tenant scenario is pretty common, and many deployments have them.
```
stefanotorresi, line 15: We're not really reverting anything, since we don't have a previous state of things. :)

```suggestion
This limitation must be overcome, as a multi-tenant scenario is pretty common, and many deployments have them.
```

stefanotorresi (commented):
I have just a minor nitpick, the rest is up to Jan's very thorough comments. :)

adr/0011-sap-system-database-aggregate-split.md
arbulu89: Not really. It means that the logic behing this connection between SAP system and database that until that was handled in the SAP system aggregate, now if handled by event handlers.
At the end, when both were 1 aggregate, you can do everything in the aggregate code, but once they are different things, "external" side effect like event handlers are required

adr/0011-sap-system-database-aggregate-split.md
```diff
@@ -0,0 +1,33 @@
...
+
+Date: 2024-04-04
+
+## Status
+
+Accepted
+
+## Context
+
+In the initial implementation of Trento, the SAP system and database states are stored in the same `sap_system.ex` aggregate. This is making the assumption that the relationship between a SAP system and a database is of 1 to 1, meaning that a database is only used by one SAP system. This is not an accurate description of the reality, as a database can handle multiple SAP systems in a multi-tenant environment.
```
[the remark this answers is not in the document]
EMaksy: What about 
> .. SAP system and a database is 1 to 1, meaning ..

instead of 

> .. SAP system and a database is **of** 1 to 1, meaning ..

adr/0011-sap-system-database-aggregate-split.md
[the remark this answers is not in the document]
arbulu89: That we have discarded using the prefix option, because if we do that, and only do that for the sap system (and we cannot do it for the others, at least by now), the SAP system stream in the event store differs from the others on how it is used.
The other streams will continue being simple uuid strings, but the sap system becomes something like `sap-system-uuid...` which makes it uglier. And as most of the people are not aware about this prefixes and there is not any easy place to document it, so we would lose a bit of how it works.

arbulu89: So nice!

arbulu89:
Thank you @janvhs and @stefanotorresi ,
Your suggestions (I think all of them), applied.

janvhs (approved):
Looks good :)

adr/0011-sap-system-database-aggregate-split.md
```diff
@@ -0,0 +1,33 @@
...
+
+Date: 2024-04-04
+
+## Status
+
+Accepted
+
+## Context
+
+In the initial implementation of Trento, the SAP system and database states are stored in the same `sap_system.ex` aggregate. This makes the assumption that the relationship between a SAP system and a database is of 1 to 1, meaning that a database is only used by one SAP system. This is not an accurate description of the reality, as a database can handle multiple SAP systems in a multi-tenant environment.
```
EMaksy, line 11: Do we need the 'of' in the 1 to 1 sentence? 

> a database is of 1 to 1 ...

Maybe just 
> a database is 1 to 1 ...

adr/0011-sap-system-database-aggregate-split.md
```diff
@@ -0,0 +1,33 @@
...
+
+In order to deal with all of these old events, besides upcasting, we have considered a couple of options when old events and IDs are mixed up. Initially, [prefixing](https://hexdocs.pm/commanded/Commanded.Commands.Router.html#module-define-aggregate-identity) the `sap_system.ex` aggregate and routing it with a `sap-system-` prefix looked a good option. It removes issues that could arise when SAP system commands with old IDs, now belonging to databases, are dispatched, as they are routed with the prefix together with the ID. We have discarded this option, as it converts the SAP system stream different from the others, and hides a bit what's going on in these scenarios. We decided against this solution, as it realistically only occours when the `DeregisterApplicationInstance` is dispatched by the deregistration process manager. Therefore, we will just handle this particular command. The command will be enriched in a middleware protocol, so commands with IDs, now belonging to databases, will be ignored.
```
EMaksy, line 25: Did you mean different as an adjective? 

> as it converts the SAP system stream different from the others

This could sound nicer, what do you think ? 

> as it converts the SAP system stream **differently** from the others

EMaksy (approved):
Hey man, lgtm, just left some small suggestions. 
Contentwise it makes sense to me, thanks for writing it down together!

rtorrero (approved):
Looks good @arbulu89, my only comment is that it would maybe help if you split the two proposals into separate paragraphs, like:
```
Proposal 1: use a prefix for the sap systems aggregate
<explain the proposal here, pros and cons>

Proposal 2: handle DeregisterApplicationInstance separately
<explain the proposal here, pros and cons>

File: adr/0011-sap-system-database-aggregate-split.md (hunk 6 above)
Comment (arbulu89) [reply adr/0011-sap-system-database-aggregate-split.md:11]: I don't know. For me both look correct. The one with `of` looks more accurate though

**Comment by arbulu89** [comment]
> Looks good @arbulu89, my only comment is that it would maybe help if you split the two proposals into separate paragraphs, like:
> 
> ```
> Proposal 1: use a prefix for the sap systems aggregate
> <explain the proposal here, pros and cons>
> 
> Proposal 2: handle DeregisterApplicationInstance separately
> <explain the proposal here, pros and cons>
> ```

I don't really one to emphasize that part, as it would stand out for the rest of the text.
At the end, it is just an additional decision we took in overall.

File: adr/0011-sap-system-database-aggregate-split.md (hunk 7 above)
Comment (arbulu89) [reply adr/0011-sap-system-database-aggregate-split.md:25]: I meant that the SAP system stream would be different compared with the others

File: adr/0011-sap-system-database-aggregate-split.md (hunk 6 above)
Comment (stefanotorresi) [reply adr/0011-sap-system-database-aggregate-split.md:11]: no actually Eugen is right, I think you're translating literally from Spanish (I guess because it would be similar in Italian) but in English you need to drop the "of".

File: adr/0011-sap-system-database-aggregate-split.md (hunk 7 above)
Comment (stefanotorresi) [inline adr/0011-sap-system-database-aggregate-split.md:25]: FIFY
```suggestion
In order to deal with all of these old events, besides upcasting, we have considered a couple of options when old events and IDs are mixed up. Initially, [prefixing](https://hexdocs.pm/commanded/Commanded.Commands.Router.html#module-define-aggregate-identity) the `sap_system.ex` aggregate and routing it with a `sap-system-` prefix looked a good option. It removes issues that could arise when SAP system commands with old IDs, now belonging to databases, are dispatched, as they are routed with the prefix together with the ID. We have discarded this option, as it would make the SAP system event stream the only one different from the others, and it would hide a bit what's going on in these scenarios. We decided against this solution, as it realistically only occours when the `DeregisterApplicationInstance` is dispatched by the deregistration process manager. Therefore, we will just handle this particular command. The command will be enriched in a middleware protocol, so commands with IDs, now belonging to databases, will be ignored.
```

stefanotorresi (changes requested):

arbulu89:
@stefanotorresi I would need a new review from you

stefanotorresi (approved):

PR #47, opened 2024-05-09
Add SUMA integration ADR

nelsonkopliku:
As per the title: SUMA integration ADR.

### DIFF
--- a/adr/0013-suma-integration.md
+++ b/adr/0013-suma-integration.md
@@ -1,0 +1,46 @@
+# 13. Tracking Software updates discovery results
+
+Date: 2024-05-07
+
+## Status
+
+Accepted
+
+## Context
+
+Trento features currently building on top of [SUSE Manager](https://www.suse.com/products/suse-manager/) (**SUMA** from here onwards) integration mainly consist in:
+- running software updates discoveries
+- exposing relevant patches and upgradable packages
+
+At the moment there is no known reactiveness in SUMA allowing Trento to be notified about state changes, so Trento proactively triggers software updates discoveries at relevant moments (ie Host registration, change in SUMA settings, in a scheduled fashion), potentially changing host health.
+
+Information used to determine a discovery outcome is the result of several requests to SUMA and is the same data useful to be exposed as relevant patches and upgradable packages.
+
+At discovery time SUMA is always queried directly in order to get a fresh representation of its state, however since the same kind information retrieved during discovery is what at the end has to be exposed as relevant patches and upgradable packages, we need to decide how the exposing side of the integration is treated.
+
+## Considered Options
+
+- **Proxied Results Retrieval**: Retrieval of software updates is proxied to SUMA
+- **Tracking Software Updates Discovery results**: Software updates discovery results are tracked in Trento at discovery time and exposed via API
+
+## Decision
+
+At the time of writing the following was taken into account as a driver for the integration
+
+> *Trento must be consistent with itself*: ie if there are patches affecting a hosts's health, Trento must provide a consistent information from different application sections, even if it means being out of sync with the actual state in SUMA
+
+As mentioned, software updates discovery results is what we rely on in order to changing the host's health, and is also the relevant information to be exposed by Trento.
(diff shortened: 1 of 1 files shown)


adr/0013-suma-integration.md
```diff
@@ -0,0 +1,141 @@
...
+
+Date: 2024-05-07
+
+## Status
+
+Accepted
+
+## Context
+
+Keeping SAP infrastructure up-to-date and secure is a crucial task for system administrators, allowing vulnerabilities reduction and overall system improvement in terms of features and performance.
```
arbulu89, line 11: To be honest, this initial chapter looks a bit "actually" out of context.
As it doesn't talk about any technical challenge.
It talks about about a product new feature.

I think the context should talk about the challenge of communicating with a 3rd party component that doesn't have any reactiveness, and we are forced to poll data in periodical basis.

adr/0013-suma-integration.md
```diff
@@ -0,0 +1,141 @@
...
+
+These premises make it both **strategically and technically convenient** leveraging SUMA's capabilities within Trento to provide a comprehensive SUSE driven experience to SAP userbase.
+
+> Current integration only focuses on enhancing visibility by running software updates discoveries. No corrective actions are taken on the hosts by Trento.
+>
+> SUMA version, and related HTTP APIs, currently targeter for integration is [4.3](https://documentation.suse.com/suma/4.3/api/suse-manager/index.html)
+
+## Considered Options
+
+- **Eventless Integration**: just proxying SUMA API calls directly within Trento's APIs, without the need of interacting with Trento's event driven nature
```
arbulu89, line 27: I think that here, on top of the table, we had 2 real scenario, as the health usage on the host health was initially taken as granted, as it was a business requirement.

1. Query and show data coming directly from SUMA (with the disadvantage that the data coming might not be consistent with the host health)
2. Query and store data coming from SUMA, so the health of the host is calculated always with the latest discovered suma info

With that being said, I would said that this initial implementation goes with the 2nd option, as we don't have any capability by now to know if there are changes in SUMA without polling directly.
With this, I would say that the 1st option could be "workarounded" just dispatching the health updated command everytime we query SUMA, but we discarded for not adding this side effect in "read" api query

arbulu89 (commented):
Hey @nelsonkopliku ,
I didn't have yet the chance to go through all of it.
I think that we should focus the ADR in the debate of "all proxied" or "proxied data save in trento DB".
Which was the part that has real impact on the decision and the real dilema.
The fact of having the health affected is a business decision so not really something we could really architect.

In my opinion, we should comment 2 options, and why we chose the 2nd by now.
We can expose as well that we do a scheduled query as SUMA doesn't have any way of notifying 3rd parties of its changes.

I think with this we would cover all the main discussions we had.
For me, personally, everything else is not really a architectural debate, because there weren't real dilemas. They were things that actually came from our Product owner in his tests (for example, that the a discovery must be made on host registered). We could list them if needed in a simple bullet list fashion, without going to complex.

That being said, I want to hear the opinion for others as well. Maybe having such a super detailed record is something they like

nelsonkopliku:
Thanks for the feedback @arbulu89!

So, yes, there may be different focus levels.

I'd be fine in re-scoping only on _track vs non-track suma results_.

I more or less unconsciously went with this exhaustiveness to give comprehensive overview to newcomers :smile:

nelsonkopliku:
Hey @arbulu89, I still think there was interesting information in the initial _extended version_, however I see that maybe not for this specific ADR. Rescoped to what suggested :+1:

rtorrero (approved):
LGTM

arbulu89 (approved):
@nelsonkopliku 
I like much much more this version, actually I find it really good.
And I appreciate the effort you put in the other condensed version, but I personally think that it was a bit out of context for a ADR.
Maybe we could put that other content in some other `doc`.

Thank you man!!

nelsonkopliku:
Thanks @arbulu89!

> Maybe we could put that other content in some other doc.
I was actually thinking about that, so I kept a copy of that version :smile:

dottorblaster:
Sorry, late to the party.

I also wanted to praise the new version of the ADR, thank you for this @nelsonkopliku :+1:

PR #52, opened 2024-08-07
Add checks decoupling ADR

dottorblaster:
Adding an ADR to sum up the state of the discussion about the checks' decoupling

### DIFF
--- a/adr/0014-decoupling-of-trento-checks-from-wanda.md
+++ b/adr/0014-decoupling-of-trento-checks-from-wanda.md
@@ -1,0 +1,31 @@
+# 14. Decoupling of Trento checks from Wanda
+
+Date: 2024-08-06
+
+## Status
+
+Accepted
+
+## Context
+
+Trento includes a component called Wanda that is responsible for performing various configuration checks on the system.
+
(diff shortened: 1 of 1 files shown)


arbulu89 (approved):
Thank you @dottorblaster 
My doubt now is how we are going to ship the currently existing checks.
Are we going to have some of them in the wanda project, or are we going to move all checks to a separated project?
And specially how we deliver the checks in a k3s setup. Is the user required to install the checks using the RPM and mount the folder in k3s?
I'm talking more about the delivery and installation than the decoupling itself.

PD: specially in cloud based k3s deployments, where the user cannot go and install RPM packages and mount volumes as they wish

dottorblaster:
@arbulu89 you're right, I was assuming too much :smile: 

The plan there is having a separate repo for all the checks, build the volume through our infrastructure and then publish the volume to our registry, so we can put that into our own Helm chart and users can download the artifact automatically.

I'll write that into the ADR, thanks for the feedback!

dottorblaster:
Also on a second thought I have to play a bit with OBS in order to get an image of our checks :grimacing: thanks for pointing that out!

dottorblaster:
@arbulu89 in the end we will have to mount a shared volume between Wanda and a sidecar container that will have to copy the checks on startup. Thanks for pointing that out, I completely missed that one.

PR #53, opened 2024-08-12
Update manual installation guide  for oidc integration

EMaksy:
This pr adds an Integration chapter for the OIDC implementation. Additionally, we can add further integrations like oauth2 and SAML as subchapters. This can become an own section in the official [Trento documentation](https://documentation.suse.com/sles-sap/trento/html/SLES-SAP-trento/index.html#sec-trento-installing-trentoserver).

We may need to discuss the structure where to put it, but we could migrate it to [chapter 4 ](https://documentation.suse.com/sles-sap/trento/html/SLES-SAP-trento/index.html#sec-trento-installing-trentoserver)
as point 4.4 Integration options with references to 4.2 systemd deployment and 4.3 Containerized deployment.

Let's see, I would suppose to finish the documentation for the other Protocols first and then take care for the migration.

@abravosuse what do you think?

### DIFF
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -247,7 +247,6 @@ AMQP_URL=amqp://trento_user:trento_user_password@localhost:5672/vhost
 DATABASE_URL=ecto://trento_user:web_password@localhost/trento
 EVENTSTORE_URL=ecto://trento_user:web_password@localhost/trento_event_store
 ENABLE_ALERTING=false
-ENABLE_OIDC=false
 CHARTS_ENABLED=true
 PROMETHEUS_URL=http://localhost:9090
 ADMIN_USER=admin
@@ -392,7 +391,6 @@ journalctl -fu trento-web
     --add-host "host.docker.internal:host-gateway" \
     -e AMQP_URL=amqp://trento_user:trento_user_password@host.docker.internal/vhost \
     -e ENABLE_ALERTING=false \
-    -e ENABLE_OIDC=false  \
     -e DATABASE_URL=ecto://trento_user:web_password@host.docker.internal/trento \
     -e EVENTSTORE_URL=ecto://trento_user:web_password@host.docker.internal/trento_event_store \
     -e PROMETHEUS_URL='http://host.docker.internal:9090' \
@@ -494,7 +492,7 @@ All roles and permissions should be managed by the newly created admin user with
 
 OIDC authentication is **disabled by default**.
 
-##### Enabling OIDC RPM
+##### Enabling OIDC when using RPM packages
 
 Provide the following environment variables to trento-web configuration, which is stored at ```/etc/trento/trento-web```  and restart the application to enable OIDC integration.
 
@@ -509,11 +507,11 @@ OIDC_BASE_URL=<<OIDC_BASE_URL>>
 OIDC_CALLBACK_URL=<<OIDC_CALLBACK_URL>>
 
 ```
-##### Enabling OIDC Docker
+##### Enabling OIDC when using Docker images
 
 Provide the following environment variables to the docker container and restart the application to enable OIDC integration.
 
-```
+```bash
 docker run -d \
 -p 4000:4000 \
 --name trento-web \
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -475,18 +475,12 @@ Expected output if Trento web/wanda is ready and the database connection is setu
 
 Trento integrates with an identity provider (IDP) that uses the OpenID Connect (OIDC) protocol to authenticate users accessing the Trento web console. Authorization for specific abilities/permissions is managed by Trento, which means that only basic user information is retrieved from the external IDP.
 
-
 #### User Roles and Authentication
 
 User authentication is entirely managed by the IDP, which is responsible for maintaining user accounts. 
 A user, who does not exist on the IDP, is unable to access the Trento web console.
-If the default admin user created during the installation process matches the authenticated user’s IDP username, that user is automatically granted `all:all` permissions within Trento.
-Create a new user on IDP and login to Trento to finalize the user creation on Trento.
-Set `all:users` permissions with the default admin user to the newly created admin user.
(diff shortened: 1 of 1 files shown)


guides/manual-installation.md
```diff
@@ -247,6 +247,7 @@ AMQP_URL=amqp://trento_user:trento_user_password@localhost:5672/vhost
 DATABASE_URL=ecto://trento_user:web_password@localhost/trento
 EVENTSTORE_URL=ecto://trento_user:web_password@localhost/trento_event_store
 ENABLE_ALERTING=false
+ENABLE_OIDC=false
```
arbulu89, line 250: You can document these variables, but they are not mandatory

guides/manual-installation.md
```diff
@@ -391,6 +392,7 @@ journalctl -fu trento-web
     --add-host "host.docker.internal:host-gateway" \
     -e AMQP_URL=amqp://trento_user:trento_user_password@host.docker.internal/vhost \
     -e ENABLE_ALERTING=false \
+    -e ENABLE_OIDC=false  \
```
arbulu89, line 395: The same, you don't really need to false, specially as you document later on how to enable it

arbulu89 (commented):
All good from my side.
I would double check with @abravosuse 
Anyway, I'm really surprised to find that our internal docs have more information than the external one. Maybe we should explain a bit more about the OIDC integration. For example, that authorizations are still managed completely by trento etc

guides/manual-installation.md
EMaksy: I removed this lines,  like you are saying, they are not mandatory, we explain it explicitly in the integration section. 
Additionally this would require to update the official documentation too, better to keep it spitted!

guides/manual-installation.md
```diff
@@ -469,6 +469,69 @@ Expected output if Trento web/wanda is ready and the database connection is setu
 {"ready":true}{"database":"pass"}
 ```
 
+## Integrations
+
+### OpenID Connect
+
+Trento integrates with an identity provider (IDP) that uses the OpenID Connect (OIDC) protocol to authenticate users accessing the Trento web console. Authorization for specific abilities/permissions is managed by Trento, which means that only basic user information is retrieved from the external IDP.
+
```
arbulu89, line 477: I would remove this not needed new line here

guides/manual-installation.md
```diff
@@ -469,6 +469,69 @@ Expected output if Trento web/wanda is ready and the database connection is setu
...
+
+#### User Roles and Authentication
+
+User authentication is entirely managed by the IDP, which is responsible for maintaining user accounts. 
+A user, who does not exist on the IDP, is unable to access the Trento web console.
+If the default admin user created during the installation process matches the authenticated user’s IDP username, that user is automatically granted `all:all` permissions within Trento.
```
arbulu89, line 483: I would clarify here that this exactly happens when the `ADMIN_USERNAME` variable matches.
Maybe like:

```
If the default admin user created during the installation process (using the `ADMIN_USER` variable, `admin` being the default value) matches the authenticated user....
```

guides/manual-installation.md
```diff
@@ -469,6 +469,69 @@ Expected output if Trento web/wanda is ready and the database connection is setu
...
+Create a new user on IDP and login to Trento to finalize the user creation on Trento.
+Set `all:users` permissions with the default admin user to the newly created admin user.
+
+All roles and permissions should be managed by the newly created admin user with the `all:users` permissions inside Trento, independent of the IDP.
+
+> **Note:** Multi-Factor Authentication cannot be enabled for the default admin user, keeping its password safe is essential
```
arbulu89, line 489: You cannot enable this either way. MFA is not usable when 3rd party IDPs are used.
Authentication is managed by the IDP so trento cannot do anything about this.

guides/manual-installation.md
```diff
@@ -469,6 +469,69 @@ Expected output if Trento web/wanda is ready and the database connection is setu
...
+
+Trento integrates with an identity provider (IDP) that uses the OpenID Connect (OIDC) protocol to authenticate users accessing the Trento web console. Authorization for specific abilities/permissions is managed by Trento, which means that only basic user information is retrieved from the external IDP.
+
+
+#### User Roles and Authentication
+
+User authentication is entirely managed by the IDP, which is responsible for maintaining user accounts. 
+A user, who does not exist on the IDP, is unable to access the Trento web console.
+If the default admin user created during the installation process matches the authenticated user’s IDP username, that user is automatically granted `all:all` permissions within Trento.
+Create a new user on IDP and login to Trento to finalize the user creation on Trento.
```
arbulu89, line 484: I don't think we should have this paragraph (from here to `Enabling OIDC` actually). In a IDP environment, they should know what users they have and how to manage them. We cannot ask them to create users.
They should only know that they must put at least one user matching the `ADMIN_USER` so they are able to change abilities for other authenticated user.

arbulu89 (commented):
Some comments

guides/manual-installation.md
```diff
@@ -469,6 +469,63 @@ Expected output if Trento web/wanda is ready and the database connection is setu
...
+
+### OpenID Connect
+
+Trento integrates with an identity provider (IDP) that uses the OpenID Connect (OIDC) protocol to authenticate users accessing the Trento web console. Authorization for specific abilities/permissions is managed by Trento, which means that only basic user information is retrieved from the external IDP.
+
+#### User Roles and Authentication
+
+User authentication is entirely managed by the IDP, which is responsible for maintaining user accounts. 
+A user, who does not exist on the IDP, is unable to access the Trento web console.
+During the installation process, a default admin user is defined using the `ADMIN_USER` variable, which defaults to `admin`. If the authenticated user’s IDP username matches this default admin user's username, that user is automatically granted `all:all` permissions within Trento.
```
arbulu89, line 482: Let's remove the `default` word from `matches this default admin users's...`

arbulu89 (approved):
Looks good

PR #54, opened 2024-08-27
Update RPM and Docker Installation Instructions for trento-checks

janvhs:
This PR updates the installation instructions to include the new `trento-checks` package.

For RPM based installations it is quite easy, since the user can just install `trento-checks`. For the Docker method there is however an open question.

How does the user manage the checks in a manual Docker install?

1. The user downloads the checks to their local file system and mounts them as a volume to the container's `/usr/trento/checks` mount point,
2. We ship an OCI artifact, if that's possible for this setup
3. The user derives from the wanda image with their own OCI image and installs the checks there.
4. Wanda preinstalls trento-checks and we instruct the user how to do 1. or 3.

What do you think @dottorblaster ?

### DIFF
--- a/guides/manual-installation.md
+++ b/guides/manual-installation.md
@@ -360,9 +360,10 @@ journalctl -fu trento-web
 1.  Install the checks on the system in a shared volume
 
     ```bash
-    docker run -it \
-    -v /usr/share/trento/checks:/usr/share/trento/checks \
-    registry.suse.com/trento/trento-checks:1.0.0
+    docker volume create trento-checks \
+      && docker run \
+      -v trento-checks:/usr/share/trento/checks \
+      registry.suse.com/trento/trento-checks:1.0.0
     ```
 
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
To me this looks good, maybe let's wait merging this until we have also the containers/kubernetes part so we can update all at once?

EMaksy:
Hey great job @janvhs!
As a comment:
+1 to what @dottorblaster  said, let's wait until we have everything together!
Why?  Well the manual installation guide must be migrated to the [official documentation](https://documentation.suse.com/sles-sap/trento/html/SLES-SAP-trento/index.html#sec-trento-installing-trentoserver), which has a different structure and it is easier to do it once then multiple times :)

EMaksy (commented):
Just saw a small thing, have a look  :+1:

dottorblaster (approved):
@janvhs just rebase this and merge :+1:

PR #62, opened 2025-01-16
Checks Customization RFC

nelsonkopliku:
Here we go: RFC for the Checks Customization feature.

TL;DR
The thing mainly boils down to where the custom values are stored (web vs wanda) and related aspects.

Here's the preview [0001-checks-customization.md](https://github.com/trento-project/docs/blob/checks-customization/rfc/0001-checks-customization.md)


rfc/0001-checks-customization.md
```diff
@@ -0,0 +1,325 @@
...
+
+# Summary
+
+Certain checks in the Trento checks catalog should be customized by customers on a target basis to adjust the corresponding expected values to ones that work better in their environment.
+
+### Use Cases outline
+
+- As a user, I want to override a Check's expected values on a target basis (host or cluster at the moment) so that it matches my specific environment needs
+- As a user, I want the overridden values to be used in a checks execution
+- As a user, I want to reset a cutomized check to default so that I can return to using SUSE's suggested best practices
```
EMaksy, line 15: Typo in: cutomized

nelsonkopliku: Fixed https://github.com/trento-project/docs/pull/62/commits/1340ca57985c71e6ba8f29db40195b740e752db5

rfc/0001-checks-customization.md
```diff
@@ -0,0 +1,325 @@
...
+### Apply custom values
+This operation allows to apply custom values for a given customizable check that has not been customized yet.
+The operation allows to apply a custom value for at least one customizable value, meaning that if a check defines more than one **expected value**, this operation allows to customize one, many or all of them (ie. if a check has 3 expected values defined, a user might be interested in customizing only one of them and use the default ones for the others)
+
+#### Endpoint
+
+`POST /checks/:check_id/values`
+```json
+{
+    "target_id": "target-uuid",
```
arbulu89, line 113: What are this `target_id` and `group_id`?

rfc/0001-checks-customization.md
```diff
@@ -0,0 +1,323 @@
...
+    "customizable": true,
+    "customized": true,
+    "custom_values": [
+      //...
+    ]
+  },
+  // other checks
+]
+
+// Note: customization related new fields could be grouped into their own entry, rathed than adding all of them at the root level
```
arbulu89, line 199: Typo `rathed` -> `rather`

arbulu89 (commented):
Hey @nelsonkopliku !
Really complete!
After reading the whole document, so initial thoughts:
- I agree that I think wanda is the place where this data should be stored. Otherwise, the attachment between web and wanda would become to big. Web should know too many details about checks and changes on what part would impact the other, making things complex
- Making a check customizable or not: if the premise is that values must be customizable depending on "human" logic (I don't want this being customizable), I don't see more options than adding a simple `customizable: boolean` new field to the `values` entries. If nothing is set, `true` being the default.
- Creating custom values: The proposal looks fine. We need to be extra careful with the type of the values to set and do proper validation. In `json` everything looks simple, in the frontend things might become nasty (anyway, most probably something easy to solve).
- Reading customized values: I think the best would be to have a new query, where you can query the customized values for each target. Yes, it would force to query the catalog and this to display things in the frontend, but I think it is an acceptable tradeoff. Otherwise, we start implementing the feature oriented to frontend visualization. Anyway, I don't have strong opinions. I think we will have a better idea once we start implementing and playing with this.
- Authorization: We already have this precedence. The checks execution button is only available for authorized users, but doesn't mean the endpoint is protected in wanda. By now, I would omit any relation to this for this epic and simply protect frontend operations. If we see that Wanda needs user based authorization, maybe we need some additional unrelated works, because it would include more things than only this.
- Logging: The same, I would only log operations done through the web/frontend. If someone uses wanda directly, well, as we don't have authorization there, so be it. Again, if in the future we include this in wanda, we can rethink again.

rfc/0001-checks-customization.md
nelsonkopliku: target_id is basically a host id and group_id is either a cluster id or a host id depending on whether we're running checks for a host or for a cluster. 

Wanda uses those.

nelsonkopliku:
thanks for the review, @arbulu89 

> Making a check customizable or not: if the premise is that values must be customizable depending on "human" logic (I don't want this being customizable), I don't see more options than adding a simple customizable: boolean new field to the values entries. If nothing is set, true being the default.
>
yes, it is a considered option. and very likely will add that.

> Reading customized values: I think the best would be to have a new query, where you can query the customized values for each target. Yes, it would force to query the catalog and this to display things in the frontend, but I think it is an acceptable tradeoff. Otherwise, we start implementing the feature oriented to frontend visualization. Anyway, I don't have strong opinions. I think we will have a better idea once we start implementing and playing with this.
>
Yes, a new query will be needed. As we get there we'll see what fits best.

> Authorization: We already have this precedence. The checks execution button is only available for authorized users, but doesn't mean the endpoint is protected in wanda. By now, I would omit any relation to this for this epic and simply protect frontend operations. If we see that Wanda needs user based authorization, maybe we need some additional unrelated works, because it would include more things than only this.
>
True and I would take the chance to improve that aspect. Only enabling disabling buttons in the frontend is not a long term solution, imho.
That said auth concerns other than frontend related will be deferred and raised again at a later stage.

> Logging: The same, I would only log operations done through the web/frontend. If someone uses wanda directly, well, as we don't have authorization there, so be it. Again, if in the future we include this in wanda, we can rethink again.
>
Logging customization activities is an AC for the feature. We can defer it, but we need to address is, eventually.

btw, added this commit this morning https://github.com/trento-project/docs/pull/62/commits/04bcc07df0c8d455cbd4348a64283593e2d4e805 just in case you did not refresh the page while reviewing 😄

rfc/0001-checks-customization.md
nelsonkopliku: nope. a customized value for a cluster is applied to all the hosts of that target

CDimonaco:
Thanks Nelson for the RFC,

```
having checks and their customization in different places makes it harder to operate when checks change (meaning that if a check changes its spec we need to react to that and possibly invalidate a previously made customization for instance. We have a similar situation with selected checks stored in web and the possibility that a selected check is removed)
``` 

As far as I remember, we already have some concepts of check execution in the web that are coupled to wanda, do you think with your plan on this feature we can use what we learn to also fix this potential issues?

Anyway no strong opinions, thanks for sharing :rocket:

nelsonkopliku:
hey @CDimonaco thank you for the review.

You're right, we have the list of the selected checks which causes issues when checks gets removed from the catalog, but they've been selected.

We can certainly leverage learnings from this feature to improve that. At due time.

EMaksy (approved):
Thanks for the sync up, for me good to start :)

PR #64, opened 2025-01-27
Agent operations framework adr

CDimonaco:
This PR includes the adr for operations framework on the agent side.


adr/0017-agent-operations-framework.md
```diff
@@ -0,0 +1,129 @@
...
+
+Following the new epics and product decisions, we need to build a framework capable of executing write operations on SAP machines.
+To achieve this, we have decided to use the agent, as it runs on the machines with root privileges, has all the necessary permissions to execute commands, and includes infrastructure to consume messages from other components of Trento.
+
+Unlike fact-gathering operations, we have chosen to delegate this new development to a dedicated library rather than embedding the code directly into the agent's codebase. This approach ensures better separation of concerns and allows the release of additional tools supporting operations without affecting the agent's codebase.
+
+The requirements for this new development are very specific:
+
+- Operations must be atomic and include rollback capabilities.
+- Operations must accept arguments to enable actions in different contexts without requiring dedicated code for each case.
```
nelsonkopliku, line 19: question: I believe I am missing some details here
> Operations must accept arguments to enable actions in different contexts without requiring dedicated code for each case
>
Could you help out with an example or point to some extra information to properly understand, please?

adr/0017-agent-operations-framework.md
```diff
@@ -0,0 +1,129 @@
...
+Below is a flowchart illustrating the transactional flow:
+
+![flow_chart](https://github.com/trento-project/workbench/raw/main/flow_chart.png)
+
+## Registry
+
+The Registry manages all available operators.
+Each operator has a version. By default, the latest version is fetched if no specific version is requested.
+
+The operator naming convention is: `<operatorname>@<version>`
```
nelsonkopliku, line 113: **question:** is the `operatorname` just the `operationID` mentioned after, or are they two different things?

nelsonkopliku (commented):
Nice, thanks! Just dropping a couple of questions for me to better understand.

adr/0017-agent-operations-framework.md
CDimonaco: Yes we can say that the operatorname is like the name of the gatherer the operationID is like the execution_id of a check execution a unique identifier for an operation

adr/0017-agent-operations-framework.md
```diff
@@ -0,0 +1,129 @@
...
+## Status
+
+Accepted
+
+## Context
+
+Following the new epics and product decisions, we need to build a framework capable of executing write operations on SAP machines.
+To achieve this, we have decided to use the agent, as it runs on the machines with root privileges, has all the necessary permissions to execute commands, and includes infrastructure to consume messages from other components of Trento.
+
+Unlike fact-gathering operations, we have chosen to delegate this new development to a dedicated library rather than embedding the code directly into the agent's codebase. This approach ensures better separation of concerns and allows the release of additional tools supporting operations without affecting the agent's codebase.
```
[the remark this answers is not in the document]
CDimonaco: Exactly is a golang library, it's like the gatherers of the checks, they are built into the code (Except there are no plugins for operators)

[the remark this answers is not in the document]
arbulu89: Yes, correct.

By the way, this part, more than the `Context` chapter is a technical decision, so I would write it in the `Decision` chapter as-is.

adr/0017-agent-operations-framework.md
arbulu89: The first operation we are working on is `saptune`. Applying saptune solutions specifically.
Different hosts need different solutions. We have `HANA`, `NETWEAVER`, `S4HANA`, `HANA=S4HANA`, etc.
So we are going to send this value as argument. This way, the operation will do a slightly different thing based on the argument.

adr/0017-agent-operations-framework.md
arbulu89: different things. The `operatorname` is a unique value that identifies the operator.
The `operationID` is the ID of the current operation execution so to say

adr/0017-agent-operations-framework.md
```diff
@@ -0,0 +1,129 @@
...
+- Operations must be versioned, with backward compatibility ensured for previous versions in the event of updates.
+- Operations must be idempotent.
+
+The agent will consume the library containing these operations and use it to fulfill operation requests from other components.
+
+## Decision
+
+The library handling operations is named [workbench](https://github.com/trento-project/workbench).
+
+## Glossary
```
arbulu89, line 30: Honestly, I don't really like this `Glossary` chapter. It is rather confusing and I think the subsequent next chapters perfectly explain all the details.
I don't think we need such a summary.
I would personally remove it

arbulu89 (commented):
Hey @CDimonaco 
Great job.
I have noted some nitpick things.
Besides that, I miss what the `operator` returns as information.
I understand that it returns the `diff` and some sort of result: if the operation went OK.
I remember that at some point we talked about returning the stage where the operator stopped.
At the end, wanda needs to know if the operation in this agent"
1. Updated something
2. Didn't update
3. Rolled back
4. Failed (so the user needs to check)

adr/0017-agent-operations-framework.md
```diff
@@ -0,0 +1,129 @@
...
+    commit(ctx context.Context) error
+    rollback(ctx context.Context) error
+    verify(ctx context.Context) error
+    operationDiff(ctx context.Context) map[string]any
+}
+```
+
+
+These methods are invoked by the Executor, which wraps the operator. All operators are exposed through a constructor function, returning operators already wrapped in an Executor.
+Executor
```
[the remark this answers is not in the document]
CDimonaco: Yes!

CDimonaco:
> Hey @CDimonaco Great job. I have noted some nitpick things. Besides that, I miss what the `operator` returns as information. I understand that it returns the `diff` and some sort of result: if the operation went OK. I remember that at some point we talked about returning the stage where the operator stopped. At the end, wanda needs to know if the operation in this agent"
> 
>     1. Updated something
> 
>     2. Didn't update
> 
>     3. Rolled back
> 
>     4. Failed (so the user needs to check)

Yep we have the failed step I forgot to mention in the doc, thx!

CDimonaco:
@arbulu89 I addressed the feedbacks, wdyt?

adr/0017-agent-operations-framework.md
```diff
@@ -0,0 +1,124 @@
...
+- Operations must be versioned, with backward compatibility ensured for previous versions in the event of updates.
+- Operations must be idempotent.
+
+The agent will consume the library containing these operations and use it to fulfill operation requests from other components.
+
+The library handling operations is named [workbench](https://github.com/trento-project/workbench).
+
+### Example:
+
+- **Operator**: `saptuneapplysolution` - Applies a SAP Tune solution using the solution name as an argument.
```
arbulu89, line 32: I don't really like this example. I think it is more confusing than helping, as both the operator and operation as mixed, and in essence both things are the same. I would have this as:
```
Find the next operator as example:

`saptuneapplysolution` is an operator that applies saptune solutions. It will apply the solution given as argument. If a solution is already applied, no action is taken. In case of an error, revert the solution specifies as an argument to return to a state where no solutions are applied.
```

nitpick: `Applies a saptune solution`, it is the tool name. Let's stick to it.
Nobody knows it as `SAP Tune` hehe
The same for the next line

arbulu89 (approved):
Perfect @CDimonaco 
Super clear to me.
I just recommended 2 additional changes. Specially the `operator/operation` distinction. I don't like it, as it mixes things. Let's simply stick to the operator term here

adr/0017-agent-operations-framework.md
```diff
@@ -0,0 +1,124 @@
...
+A write operation can either succeed or fail. Upon success, it generates a diff showing changes made during the commit phase.
+
+The operator accepts arguments in the form of a `map[string]any` to specify operation parameters. Each operator is responsible for extracting and validating these arguments.
+
+The operator follows a transactional workflow, which includes the following distinct phases:
+
+- **PLAN**
+- **COMMIT**
+- **VERIFY**
+- **ROLLBACK**
```
balanza, line 47: This applies to a single write operation, right? If an operator executes multiple write operations, will they be in the same transaction?

Also, so far we are not considering distributed transactions across multiple operators in the same operations, correct?

balanza (commented):
Great job guys. That's a great feature to be added to the project.

As a little feedback, I struggle a bit with naming: operations, operator, write operations, operationId, operator name... just to name a few. Furthermore, such terms are common in the daily jargon and they might be easily misused and misunderstood.
I'd rather have a glossary for such terms to avoid ambiguities, or to choose less-ambigue names.

adr/0017-agent-operations-framework.md
```diff
@@ -0,0 +1,124 @@
...
+
+Unlike fact-gathering operations, we have chosen to delegate this new development to a dedicated library rather than embedding the code directly into the agent's codebase. This approach ensures better separation of concerns and allows the release of additional tools supporting operations without affecting the agent's codebase.
+
+The requirements for this new development are very specific:
+
+- Operations must be atomic and include rollback capabilities.
+- Operations must accept arguments to enable actions in different contexts without requiring dedicated code for each case.
+- Operations must be transactional, including distinct steps for prerequisites verification, commit, rollback, and validation of changes applied during the commit phase.
+- Operations must be versioned, with backward compatibility ensured for previous versions in the event of updates.
+- Operations must be idempotent.
```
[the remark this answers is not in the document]
arbulu89: This is true. I would rename this to `Operator`.
At the end, the agent is all about operators. I would try to avoid any `Operation` wording.
At the end, and operation is a set of steps. And each step requests a operator execution in the agents.
The agents, by know doesn't have the knowledge of the `Operation` as a whole. It is only responsible of knowing and executing this small pieces of code now called `Operators`

adr/0017-agent-operations-framework.md
arbulu89: > This applies to a single write operation, right? If an operator executes multiple write operations, will they be in the same transaction?

I guess one operator can do more than one "write" operation in the same transaction, so we can rollback the whole process. The example is the corosync file change and reload. It is the same like re-configuring nginx (easier example). You change the nginx configuration file and restart the daemon. If the restart fails, you restore the initial config file and restart. I guess we could include all this in the same operator. Otherwise things start getting complex. We didn't have yet the chance to implement such a thing though, so this can change.
 
> Also, so far we are not considering distributed transactions across multiple operators in the same operations, correct?

damn, now I feel you pair with the terminology XD
The operator shouldn't never know that there are other operator executions in the same agent or operation. That's the meaning of "atomic" I guess. The operator only knows to do one single thing deterministically.
That's the difficulty on implementing complex multi-step operations and their rollbacks. That rollbacking operations with multiple steps is difficult. That's why we are postponing this implementation and design

EMaksy:
Just as a note for the history, took the markdown source, converted it to adoc, in case any content requires change, we can address it in a follow up up

PR #65, opened 2025-01-28
Add agent operations orchestrato ADR

arbulu89:
ADR describing the agent operations orchestrator in wanda side.

To see the rendered page look at: https://github.com/trento-project/docs/blob/adr-operations-orchestrator/adr/0018-agent-operations-orchestration.md


nelsonkopliku (approved):

adr/0018-agent-operations-orchestration.md
```diff
@@ -0,0 +1,100 @@
...
+
+We have chosen to use Elixir [Dynamic Supervisors](https://hexdocs.pm/elixir/1.13/DynamicSupervisor.html) to start and handle the operation requests dynamically. This means that the operations are short-lived processes responsible for handling individual operation requests. We already do the same with check executions with a successful outcome.
+
+The process is able to:
+- Predicate if the operation should be executed in target agents
+- Dispatch the individual operator requests to the target agents
+- Wait for the reports and compute the overall step result after receiving the feedback from all the targets
+- Move to the next operation step once the previous one is successfully completed
+- During all this time, the state of the operation is saved in the database in order to have intermediate states. This can be used to report progress to other components like Web or rehydrate and restart crashed operations (this is out of cope by now)
```
balanza, line 28: **nitpick:** ‏I'd rather focus on the feature than on the implementation. I propose a rephrase:
```suggestion
- The state of the operation is maintained in case of unavailability of Wanda
```

adr/0018-agent-operations-orchestration.md
```diff
@@ -0,0 +1,100 @@
...
+As commented, the steps define the operator to be executed in the agents. Check the [Agent related ADR](https://github.com/trento-project/docs/blob/main/adr/0017-agent-operations-framework.md) for more information.
+The `required_args` field defines the arguments that must be provided from the outside with a key/value format. For example: `saptune_solution=HANA`. They will be passed to the agent operators so they can be executed.
+
+### Orchestration state machine
+
+The orchestration state machine is pretty simple. It is composed only by 2 states:
+- Dispatch the step operations to agents
+- Wait until all the agents have reported back
+
+This image represents the states and transitions:
```
balanza, line 68: #

**issue:** ‏It's actually a flowchart, not a state machine

adr/0018-agent-operations-orchestration.md
```diff
@@ -0,0 +1,100 @@
...
+
+![state_machine](images/0018_orchestration_state_machine.png)
```
balanza, line 70: **suggestion:** I'd add the predicate evaluation step for clearness.

(also, consider using [Mermaid](https://github.blog/developer-skills/github/include-diagrams-markdown-files-mermaid/) for charts as it can be included in markdown seamlessly).

adr/0018-agent-operations-orchestration.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Predicate
+
+Deciding if a step operation must be executed in some targets or not is something really useful. This will let us define some multi-agent operations where some step operations are only executed in certain nodes. Some examples could be:
+- Run cluster operations only in DC nodes
+- Run SAP operations only in nodes with specific features, like ASCS or PAS nodes
+- Exclude running operations in nodes like majority makers
+
+The predicate will be a simple RHAI expression that returns a boolean. We will pass some target specific values in the requests, so each target has its own characteristics. With these values, we will run the RHAI evaluation and decide if the operation step must be executed in this target or not.
```
balanza, line 79: **suggestion:** ‏It might be worth being explicit on:
* that RHAI expressions are executed on Wanda, not on agents
* how the agents' arguments are collected and registered.


**nitpick(language):** Also, I suggest to use _target_ ‏only to refer to agents that are actually interested by the step (hence, the predicate was confirmed).

```suggestion
The predicate will be a simple RHAI expression that returns a boolean. We will pass some agent specific values in the requests, so each agent has its own characteristics. With these values, we will run the RHAI evaluation and decide if the operation step must be executed in this agent or not.
```

adr/0018-agent-operations-orchestration.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Operations registry
+
+The operations to be executed will be stored in a registry. Unlike the checks executions, where the checks are implemented using YAML files and a specific DSL, in this case the operations will be hardcoded in the codebase, and there won't be any capability to upload them after releasing the code. We want to have control over want to implement, as the operations are really sensitive actions that could lead to malfunctions in the system if implemented incorrectly. The registry simply implements a basic map with some unique operation identifiers pointing to operations described in the [Operation](#operation) chapter.
```
balanza, line 83: **question:** ‏

> he operations will be hardcoded in the codebase

Which codebase? Wanda, Agent, Workbench...

adr/0018-agent-operations-orchestration.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Out of scope in the first implementation
+
+To make the delivery of this feature easier and faster, we have decided to let out some of the initial features we wanted to implement:
+- Complex rollback of operations. When the operation implements a lot of steps, returning to the original state in case of failure is pretty difficult
```
balanza, line 88: **issue(unambiguous language):** We  ‏should be more specific on that. What is considered `Complex`? How many is `a lot of steps`? 
If the purpose is to set the right expectations, I'd leave out any personal interpretations of that.

adr/0018-agent-operations-orchestration.md
```diff
@@ -0,0 +1,100 @@
...
+- Restart of crashed operations. We will "let it fail" by now, as in the majority of the cases, the error will be caused for things that will hardly be fixed in a subsequent execution. These can be things like: database access error, error publishing or reading from RabbitMQ, agent reporting an incorrect payload, etc
```
balanza, line 89: **question:** ‏By `crashed operations` we mean operations whose state is _somehow_ corrupted and cannot be restored?

balanza (changes requested):
Thanks @arbulu89, that goes far beyond a decision record, as it gives a lot of context on what will be put in place.

I left some comments, most of them are about wording; even though, I think being more precise will help us in the future.

adr/0018-agent-operations-orchestration.md
arbulu89: I'm going to rephrase this to something else, as the main purpose of saving is just to have the historic record of operations: 
`Save the intermediate and final state of the operation as an historic record`

adr/0018-agent-operations-orchestration.md
arbulu89: Ok, changed.  I thought the RHAI part was obvious :sweat_smile:

adr/0018-agent-operations-orchestration.md
balanza: Ok. You see, I got confused about the feature

adr/0018-agent-operations-orchestration.md
arbulu89: Not necessarily but it is part of the scenarios. It means any operation that raises and unknown error so the code fails
```
database access error, error publishing or reading from RabbitMQ, agent reporting an incorrect payload, etc
```
Rephrased

adr/0018-agent-operations-orchestration.md
balanza: So we can have a rollback of single-spet operations, that spans over multiple agents?

adr/0018-agent-operations-orchestration.md
arbulu89: Let's see what I can do

adr/0018-agent-operations-orchestration.md
arbulu89: That's inherently achieved in the agents, as they implement the operator specific rollback.
I want to express that multi-step operations rollback will not be implemented initially

arbulu89: Changed to:
```
Multi-step operations rollback. When the operation implements a lot of steps, returning to the original state in case of failure has some challenges as all intermediate steps must be rolled back in reverse other.
```

adr/0018-agent-operations-orchestration.md
arbulu89: Done!
It looks pretty neat!
Thank you for the recommendation

arbulu89:
Hey @balanza 
I think I have applied all your suggestions. Can you have a new look?

balanza (approved):
Thanks @arbulu89!

adr/0018-agent-operations-orchestration.md
balanza: Will we have a _retry_ for transient failures, such as network error?

arbulu89: My plan is **not** to, as there is not way to way I know about where you can restart the gen server only on specific errors. And even if it was, i don't know if it is possible to have a finite list of errors that we want to restart, and other where we don't want to

balanza: Don't you think the system will be too fragile? Any _glitch_ in the Wanda instance will result in an incomplete/indeterminate operation state.

EMaksy (approved):
Hey man good job!
Very informative.

adr/0018-agent-operations-orchestration.md
arbulu89: @balanza 
Yes, it won't be 100% reliable for sure. We accept that as trade-off.
I have added this commit to add additional context about the decision: https://github.com/trento-project/docs/pull/65/commits/6334257d6c47b852bb9504df5ae79d48c2c9dd8a

PR #71, opened 2025-05-06
Checks Customization ADR

nelsonkopliku:
Adding Checks Customization ADR

### DIFF
--- a/adr/0020-checks-customization.md
+++ b/adr/0020-checks-customization.md
@@ -1,0 +1,80 @@
+# 19. Checks Customization
+
+Date: 2025-04-30
+
+## Status
+
+Accepted
+
+## Context
+
+Trento provides a catalog of checks with predefined expected values, often based on SUSE's best practices. However, specific customer environments may necessitate deviations from these defaults for certain checks to accurately reflect their operational reality or specific configurations.
+
+Therefore, a mechanism is needed to 
+- allow users to override the built-in expected values of specific checks for particular targets
+- have these overrides used during check executions
+- track changes in the activity log
+- allow to opt-out from customizability for either entire checks or specific values within a check
+
+## Considered Options
+
+- **Storing Customizations in Web**
+    *   places check-related data outside the Check Engine, making standalone engine usage clunky
+    *   requires passing custom values for every execution, complicating the execution request process (message contracts/API endpoints)
+    *   managing consistency between check definitions (in Wanda) and customizations (in Web) becomes more complex, especially when check specifications change or checks are added/removed
+    *   logging customizations is simple as the activity logging subsystem is embedded in web
+- **Storing Customizations in Wanda**
+    *   keeps check-related logic and data consolidated within the Check Engine
+    *   allows for a more straightforward execution process, as Wanda can directly access its state for custom values
+    *   managing consistency upon check specifications change or checks are added/removed is simpler because all the needed information is in the same place
+    *   logging customizations needs the activity logging subsystem to be extended to support message based activities
+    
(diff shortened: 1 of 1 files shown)


adr/0020-checks-customization.md
```diff
@@ -0,0 +1,80 @@
...
+- track changes in the activity log
+- allow to opt-out from customizability for either entire checks or specific values within a check
+
+## Considered Options
+
+- **Storing Customizations in Web**
+    *   places check-related data outside the Check Engine, making standalone engine usage clunky
+    *   requires passing custom values for every execution, complicating the execution request process (message contracts/API endpoints)
+    *   managing consistency between check definitions (in Wanda) and customizations (in Web) becomes more complex, especially when check specifications change or checks are added/removed
+    *   logging customizations is simple as the activity logging subsystem is embedded in web
```
arbulu89, line 25: I would add the keeps `wanda` stateless and without need of a database usage (it would only we used to store historic records, not for execution itself). This was the major benefit of storing the customizations in web in my opinion

arbulu89 (approved):
thank you

adr/0020-checks-customization.md
nelsonkopliku: I am not sure I understand what you could add 😅 . Wanda is not stateless.

arbulu89: I mean persistent state. It didn't need a database to work (except we store the historical record of executions, what this comes as a side-effect. The execution would still work without it).
Until now, the execution was executed with data coming in the request. Now, it needs to check the database to get additional data, which I consider the state.
Making a simil, it is like functional programming. Until now, same input in the request guaranteed same execution request in the targets. Now, same input requests might have different execution requests in the target depending on what we have in the database.

nelsonkopliku: I see now, here's the commit that addresses it https://github.com/trento-project/docs/pull/71/commits/ebd9f881e61029e156319f0f307cadb1760db5c0

I understand the parallelism with functional programming, however I see it a bit of a stretchy fit in the context 🙈
We're dealing with systems of components, and necessarily with state, possibly distributed.
Not all the golden rules of fp might apply at this level IMHO 🤷 

In fact I imagine a world where upon checks selection the group-targets-selection relation (ids would be sufficient) ends up in wanda as well so that an execution can be started by just telling wanda: hey start an execution for this group.

But this is out of scope eheheh

arbulu89: Of course, they are different approaches with their pros/cons.
I personally loved the way the executions worked without relaying in the database. The same with the checks selection usage that is implemented currently. I personally really like how it works. I don't need to care to know what's in the database. I don't even care if there is a database hehe
I just know the request I want to do. One single request.

I just wanted to point that out in the document that's all (and with the added phrase is good enough).
We made the decision to go in the opposite direction and that's ok.

adr/0020-checks-customization.md
```diff
@@ -0,0 +1,80 @@
...
+- keep cohesive and consistent responsibility about check related actions and data
+- keep execution process simple and straightforward also in a standalone engine usage
+- allow for simpler management of check specifications changes and check additions/removals
+
+### Checks Customizability
+A check is deemed "customizable" if its definition includes a non-empty `values` section. 
+
+Checks using hardcoded expectation values will need refactoring to use the `values` structure to become customizable. 
+
+Certain checks may be explicitly excluded from customization as well as specific values within a generally customizable check.
```
EMaksy, line 46: I miss a bit the "why"
Why are some checks excluded?

EMaksy (approved):
Hey LGTM, left just a minor comment :)

PR #81, opened 2025-07-14
Convert Markdown documentation to Adoc

EMaksy:
This pr is a file migration from markdown to adoc, also took the chance to restructure and make the structure a bit more flat so it is easier to navigate through the documentation.

I will use this as base for the antora page and later the migration of official suse documentation and to play around with versioning

Note: some links will work after the merge, tested it manually and on the antora page


EMaksy:
Thanks @nelsonkopliku you are right the list is broken, i will fix it in the next commit!

PR #102, opened 2025-10-28
Update docs to use new wanda location in nginx and OAS support

arbulu89:
Update docs to explain the usage of some of the new values and new nginx configuration.

The changes are:
- Change nginx configuration to redirect `/wanda` requests to wanda service/container. I have taken advantage and update the whole configuration to look like what we have in ansible, which is what we are testing all the time.
- Add new `OAS_SERVER_URL` variables usage to have a usable OAS swagger and spec
- Update `trento-web.trentoWebOrigin` by `global.trentoWeb.origin`, as the value is used by more charts now

@abravosuse putting you on review as this changes affect the whole functionality, the `helm` new usage specially (you might need to update some of your scripts).

PD: the `trento-web.trentoWebOrigin` is still somehow functional, as it was before, but if this deprecated way is used the OAS schema won't work properly, and this is required to use the mcp server.

### DIFF
--- a/trento/adoc/trento-container-install.adoc
+++ b/trento/adoc/trento-container-install.adoc
@@ -144,7 +144,7 @@ docker run -d \
 -e ADMIN_PASSWORD='test1234' \
 -e ENABLE_API_KEY='true' \
 -e TRENTO_WEB_ORIGIN='trento.example.com' \
--e CHECKS_SERVICE_URL=/wanda \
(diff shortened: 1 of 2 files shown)


EMaksy (approved):
For me it looks good, only need to test i guess :+1: ,  i would say wait with merge until the others had a quick review but it is good to go :)

abravosuse (approved):

PR #112, opened 2025-11-18
Add trento installation automation docs to Antora

vicenteqa:
Add repo and *.adoc location to include it in the developer documentation.


trento-docs-site/antora.yml
```diff
@@ -125,6 +126,11 @@ ext:
       - dir: trento-docs-site/build/tmp_components/tlint/docs
         files: ['**/*.adoc', '!README.adoc']
         into: modules/tlint/pages
+    ## Trento trento-installation-automation https://github.com/trento-project/trento-installation-automation
```
nelsonkopliku:
```suggestion
    ## Trento installation automation https://github.com/trento-project/trento-installation-automation
```
---
**thought:** I am wondering whether `installation automation` could be misleading.

From my understanding the project is about provisioning the infrastructure needed to test trento, and  certainly installing the needed trento artifacts via ansible.

Let's maybe think if another name revolving around testing would fit better.

nelsonkopliku (approved):
Looking good!

PR #117, opened 2025-11-26
Operations

abravosuse:
Documentation about the first set of operation use cases that will be introduced with version 3.0.

### DIFF
--- a/trento/adoc/trento-operations.adoc
+++ b/trento/adoc/trento-operations.adoc
@@ -3,162 +3,172 @@ include::product-attributes.adoc[]
 
 == Operations
 
-Operations in {trento} are ruled by four principles:
+Operations in {trento} follow four principles:
 
-. They are protected by permissions: only users with `operation:all` permissions are allowed to perform operations.
-. The path in the UI to request an operation is a contextual one. For example, if you want to turn maintenance on in a cluster, the path that you will follow to request the operation will depend on the exact target. If you want to turn maintenance on in the entire cluster, you will request the operation from the Operation button at the top of the cluster details view. If you want to turn maintenance on in a particular node, you will do it from the burger menu in the right hand side of the corresponding node row. And if you want to turn maintenance on in a particular resource, you will do it from the burger menu in the right hand side of the corresponding resource row.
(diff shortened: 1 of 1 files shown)


abravosuse:
> Hey @abravosuse , Content wise I find it good. I didn't find any wrong policy or parameter usage.
> 
> The typical comment about using `future` form. We usually don't use it in the docs, so forms like `will/would` usually are discouraged. I would say the same with the `you` usage. I find it strange.
> 
> Anyway, I'm leaving this things to review by the docs team.

@arbulu89 Correct. Once we have reviewed the content from a technical perspective, I will ask Dima for a review and he will address all these topics.

trento/adoc/trento-operations.adoc
```diff
@@ -0,0 +1,163 @@
...
+*** Internal policies: not applicable
+*** Internal timeout: 5m
+** Enable {pace} at boot: Enabled when the service is disabled at boot in the corresponding host.
+*** Activity type: Operation Requested on a cluster host
+*** Internal policies: not applicable
+*** Internal timeout: 5m
+** Disable {pace} at boot: Enabled when the service is enabled at boot in the corresponding host.
+*** Activity type: Operation Requested on a cluster host
+*** Internal policies: not applicable
+*** Internal timeout:: 5m
```
nelsonkopliku: There actually is an internal policy for these, which are the mentioned conditions about when the operation is enabled
- Enable pacemaker: the service is disabled at boot in the corresponding host
- Disable pacemaker: the service is enabled at boot in the corresponding host

I am fine with the content, just wanted to highlight that from an API perspective you could still request the operation, but the policy will take place.

From a UI perspective this is the button being enabled or not.

trento/adoc/trento-operations.adoc
```diff
@@ -0,0 +1,163 @@
+include::product-attributes.adoc[]
+
```
EMaksy: Please add a revdate like here 
https://github.com/trento-project/docs/blob/main/trento/adoc/checks_customization.adoc?plain=1#L4

We may use this variables in the future to to show the date of an article, but at least good for us to know when writing :)

EMaksy:
Fine for me @abravosuse , feel free to set the pr from draft to review :)

EMaksy (approved):
Hey @abravosuse nicely done, nothing really to add here

PR #126, opened 2025-12-22
Grafana Alloy integration

balanza:
This RFC should serve as a blueprint for integrating system metrics with Prometheus on Grafana Alloy-based system.

The integration is needed because SLES >= 16 does not ship `node_expoter` anymore; instead, the default collector tool is Grafana Alloy.

The main goal of the RFC is to reach **feature parity** with the current solution both in terms of collected metrics and user experience.

### DIFF
--- a/content/rfc/0003-grafana-alloy-integration.adoc
+++ b/content/rfc/0003-grafana-alloy-integration.adoc
@@ -678,10 +678,11 @@ WARNING: Running without authentication in production is not recommended. Anyone
 
 ==== Trento Out-of-the-Box Installations
 
-For Trento's provided installation methods, bearer token authentication is pre-configured:
+For Trento's provided installation methods, we want to provide a sane default configuration that does not add unwanted operational effort for the average user, while allowing advanced users to customize the setup as needed.
 
-* **systemd deployments**: nginx is configured with bearer token validation
-* **Helm deployments**: Kubernetes Ingress is configured with bearer token validation
+We will provide a default configuration where bearer token authentication in place on both ansible playbook and helm chart installation scripts.
+
+For both scenarios we configure the reverse proxy to validate bearer tokens for the Prometheus remote write endpoint. The token is meant to be long-lived and static, as it is used by multiple agents.
 
 == Handling Edge Cases
 


content/rfc/0003-grafana-alloy-integration.adoc
```diff
@@ -0,0 +1,364 @@
...
+install -D -m 0644 packaging/systemd/trento-agent-alloy.service \
+    %{buildroot}%{_unitdir}/trento-agent.service
+%else
+# SLES 15 and earlier: use node_exporter-based service
+install -D -m 0644 packaging/systemd/trento-agent.service \
+    %{buildroot}%{_unitdir}/trento-agent.service
+%endif
+----
+
+NOTE: The `sle_version` macro is provided by OBS and contains the SLES version as an integer (e.g., `150000` for SLES 15, `160000` for SLES 16).
```
nelsonkopliku: **praise:** great finding! Would you mind sharing the reference to where this is documented please?

content/rfc/0003-grafana-alloy-integration.adoc
```diff
@@ -0,0 +1,909 @@
+= RFC: SLES 16 System Metrics Support via Grafana Alloy
+:author: Trento Engineering Team
+:revdate: December 2025
+:revnumber: 2.2
+:toc: left
+:toclevels: 3
```
nelsonkopliku: **nit:** not sure the toc is strictly necessary. Usually the rendering tools already generate it (github, antora)

content/rfc/0003-grafana-alloy-integration.adoc
```diff
@@ -0,0 +1,774 @@
...
+
+# Generate the Alloy configuration
+/usr/bin/trento-agent generate alloy > "${ALLOY_CONFIG_FILE}"
+
+# Set ownership and permissions
+chown "${ALLOY_USER}:${ALLOY_GROUP}" "${ALLOY_CONFIG_FILE}"
+chmod 640 "${ALLOY_CONFIG_FILE}"
+
+# Reload Alloy to pick up the new configuration
+systemctl reload alloy
```
nelsonkopliku: **issue:** if the `configure-alloy.sh` script is going to be executed in `ExecStartPre` we need to be mindful about the fact that alloy service might not be available and `systemctl reload alloy` would fail, right?

nelsonkopliku (commented):
@balanza thanks a lot for this! Great job!

I left some comments, the most concerning one is about reachability and authentication of the prometheus write url

content/rfc/0003-grafana-alloy-integration.adoc
```diff
@@ -0,0 +1,909 @@
...
+----
+%files
+%{_bindir}/%{binaryname}
+%{_unitdir}/trento-agent.service
+%if 0%{?sle_version} >= 160000
+%{_datadir}/trento/configure-alloy.sh
+%endif
+----
+
+NOTE: The `sle_version` macro is provided by OBS and contains the SLES version as an integer (e.g., `150000` for SLES 15, `160000` for SLES 16). The `configure-alloy.sh` script is only installed on SLES 16 and later.
```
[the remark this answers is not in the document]
balanza: It's defined in https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto (and actually for SLES 16 we must use `suse_version` instead)

content/rfc/0003-grafana-alloy-integration.adoc
balanza: True. We can add a `-` to make the unit to ignore failures:

```
ExecStartPre=-/usr/share/trento/configure-alloy.sh
```

See https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html

Of course, Trento might not collect data in such case, and the user will have to trobleshoot the problem (which can happen at every moment after start, too)

content/rfc/0003-grafana-alloy-integration.adoc
```diff
@@ -0,0 +1,836 @@
...
+* Development/testing environments
+* Fully isolated networks with no external access
+
+WARNING: Running without authentication in production is not recommended. Anyone with network access could push arbitrary metrics to Prometheus.
+
+==== Trento Out-of-the-Box Installations
+
+For Trento's provided installation methods, bearer token authentication is pre-configured:
+
+* **systemd deployments**: nginx is configured with bearer token validation
```
nelsonkopliku: **question:** you mean using something like https://nginx.org/en/docs/http/ngx_http_auth_jwt_module.html?

content/rfc/0003-grafana-alloy-integration.adoc
```diff
@@ -0,0 +1,774 @@
...
+[source,yaml]
+----
+# /etc/trento/agent.yaml
+
+api-key: <api-key-generated-from-the-server>
+server-url: https://localhost
+facts-service-url: amqp://guest:guest@localhost:5672
+
+# Alloy configuration options
+prometheus-url: https://trento.example.com/api/v1/write
```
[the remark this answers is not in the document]
nelsonkopliku: Alright, thanks. Almost everything resonates with me. I will follow up in the `Authorization` section for more comments.

PR #164, opened 2026-03-30
AI Integration RFC

nelsonkopliku:
Add AI integration RFC.

Visual [here](https://github.com/trento-project/docs/blob/ai-assistant-rfc/content/rfc/0004-ai-agent-integration.adoc)

### DIFF
--- a/content/rfc/0004-ai-agent-integration.adoc
+++ b/content/rfc/0004-ai-agent-integration.adoc
@@ -2,7 +2,7 @@
 
 [width="100%",cols="<18%,<82%",]
 |===
-|Feature Name |AI assistant chatbox
+|Feature Name |AI assistant
 |Start Date |Jan 1st, 2026
 |Category |AI, Architecture
 |PR |https://github.com/trento-project/docs/pull/164[#164]
@@ -134,7 +134,9 @@ PROs:
 
 * Mature and feature-rich framework with a large community and ecosystem.
 * Extensive documentation and resources available.
+
 CONs:
+
 * Requires a separate deployable component
 * the Python version would require a new technology stack for the backend
 
@@ -215,7 +217,7 @@ The problem with another artifact is not about releasing it: that's being stream
 
 ==== The proposed path forward
 
-Given the above, the proposed path forward is to avoid adding a separate artifact and instead integrate the AI agent within an existing component, specifically Trento Web (Elixir implementation). 
+Given the above, the proposed path forward is to avoid adding a separate artifact and instead integrate the AI agent within an existing component, specifically Trento Web (Native Elixir implementation). 
 
 This approach minimizes architectural complexity and operational overhead allowing us to focus on the core features, which have their degree of inherent complexity.
 
--- a/content/rfc/0004-ai-agent-integration.adoc
+++ b/content/rfc/0004-ai-agent-integration.adoc
@@ -188,7 +188,9 @@ Elixir:
 * https://github.com/trento-project/web/tree/liz-testing-langchain-elixir
 * https://github.com/trento-project/web/tree/ai-native-poc
 
-Golang: no reference available, but it has been attempted.
+Golang: 
+
+* https://github.com/trento-project/mcp-server/tree/TRNT-4140-liz-alt
 
 ===== How to choose which way to go?
 
--- a/content/rfc/0004-ai-agent-integration.adoc
+++ b/content/rfc/0004-ai-agent-integration.adoc
@@ -62,6 +62,10 @@ There are 3 main areas involved in the "which technology/standard to use" questi
 * link:#_ui_to_agent_communication[UI to Agent Communication]: the protocols and technologies that enable the frontend UI to interact with the AI agent backend
 * link:#_ui-components[UI Components]: the actual implementation of the assistant widget in the frontend
 
+Additionally to these there is a fourth area around the tools that the AI agent can use, how they're provided to the agent, and how they interact with the rest of the system.
+
+* link:#_tools_integration[Tools Integration]: the way the AI agent can leverage tools
+
 The first one, Agentic Framework, drives the decision about whether we need/want to introduce a new artifact in the picture of our architecture, which at this stage is the main question we want to answer.
 
 === Agentic Frameworks
@@ -288,6 +292,80 @@ Considering:
 * branding/watermark issues
 
 The proposed direction is to defer commitment to a specific AG-UI (or non-AG-UI) compliant component library.
+
+=== Tools Integration
+
+Currently the MCP server exposes tools to AI assistants from Trento's API specifications. This is necessary for external assistants like VSCode, Claude, etc...
+
+With a natively integrated AI Agent, we have two options:
+
+* keep using MCP Server to add tools the AI Agent
+* use internal functions, where possible, as tools for AI agent
+
+==== Option 1: Keep using MCP Server to add tools the AI Agent
+
+Using the MCP server as the main "tools provider" means registering it in the AI Agent, effectively requiring to have an MCP client for it.
+
+PROs:
+
+* leverage the work already done in the MCP server to expose tools
+* any new endpoint tagged with "MCP" will be automatically available to the AI Agent as a tool
+
+CONs:
+
+* latency/network overhead, as it would require an unpredictable amount of network calls (See following note)
+* how to deal with tools only relevant for the AI Agent but that we might not need/want to expose as an API?
+
+Note on latency/networking overhead
+
+Let's consider a basic use case where a user prompt resolves in 1 API to be called.
+
+What would be the flow? (Let's consider web == AI Agent)
+
+1. User sends a prompt from the UI to the AI Agent (client -> web [not counted])
+2. AI Agent calls, at least once, the MCP server to get the list of the available tools (APIs) (web -> MCP server [1 call at least])
+3. AI Agent calls the LLM with the user prompt and the list of tools (web -> LLM provider [1 call])
+4. AI Agent receives instructions from the LLM on which tool to use, then calls the tool which is an API exposed by Trento 
+    a. if the tool is a web API (web -> MCP server -> web [2 calls])
+    b. If the tool is a Wanda  API, then there is also Token introspection involved (web -> MCP server -> wanda -> token introspection in web [3 calls])
+
+4 requests when the tool is a web API, 5 when the tool is a Wanda API. Average of 4 considering the amount of wanda tools is less than the amount of web tools.
+
+It is worth mentioning that:
+
+* there is also the roundtrip of the responses not explicitly mentioned above
+* the amount of tools to be called is unpredictable, meaning that the flow could be even more complex than the one described above and the failing points and latency could be more significant
+* authnz is re-executed many times against web when its APIs are called as tools
+
+==== Option 2: Internal tools
+
+PROs:
+
+* one less component to install and still have access to the same AI capabilities
+* opens the possiblity to expose internal functions as tools for the AI agent without necessarily exposing them as APIs, if not needed
+* less latency/networking overhead, as the calls to the tools would be mainly internal function calls instead of network calls to the MCP server
+* maximised authnz reuse within the same conversation context
+
+CONs:
+
+* exposing internal features/functions as tools for the AI agent might lead to repetition of code that is already wired up in how our controllers work, or that needs to be wired up both as a controller action and as an AI agent tool
+* since wanda is a separate component, we would need to call it as an external API (similarly to what we do with Prometheus, for instance)
+
+What would be the flow in this case? (Let's consider web == AI Agent)
+
+1. User sends a prompt from the UI to the AI Agent (client -> web [not counted])
+2. AI Agent calls the LLM with the user prompt and the list of tools (web -> LLM provider [1 call])
+3. AI Agent receives instructions from the LLM on which tool to use, then:
+    a. if the tool is a web functionality call the internal tool, no need to go outside
+    b. If the tool is a Wanda functionality, then call the related API (web -> wanda -> token introspection in web [2 call])
+
+1 request when the tool is internal to web, 3 when the tool is a Wanda API. Average of 2.
+
+The amount of network calls could be reduced by around up to 50%, however the unpredictability still remains, based on what the user prompts and what the LLM instructions are.
+
+==== The proposed solution
+
+TBD
 
 == Summary
 
--- a/content/rfc/0004-ai-agent-integration.adoc
+++ b/content/rfc/0004-ai-agent-integration.adoc
@@ -1,0 +1,295 @@
+== 4. AI Assistant Chatbox
+
+[width="100%",cols="<18%,<82%",]
+|===
+|Feature Name |AI assistant chatbox
+|Start Date |Jan 1st, 2026
+|Category |AI, Architecture
+|PR |https://github.com/trento-project/docs/pull/164[#164]
+|===
+
+== Summary
+
+Build an **AI assistant widget**, fully integrated within Trento product experience.
+
+== Motivation
+
+Currently users have only access to https://github.com/trento-project/mcp-server[Trento MCP server] to leverage AI capabilities. 
+While already quite powerful, it requires to use external tools, like VSCode, to actually interact with the assistant.
+
+We want to make AI capabilities more accessible and integrated within Trento to:
+
+* reduce the barrier of setup
+* enrich the LLM capabilities with product-specific context (UI context, RAG, etc.)
+* enhance the overall user experience
+
+This RFC addresses the development of such an integrated AI assistant by defining what it entails in overall Trento Architecture:
+
+* technologies for the agentic framework
+* communication protocols between the AI Agent and the UI
+* where data and operations belong (that is: another artifact or not?)
+
+=== Use Cases Outline
+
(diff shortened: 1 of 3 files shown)


arbulu89 (commented):
Hey @nelsonkopliku ,
Nice stuff!
I have not been involved in the topic, and most probably my input is not the most relevant one, but after reading the docs I learned some interesting things and exposes the feature nicely. Well done!

I have just a question:
If we had the chance the release some sort of "tech-preview" release, without any required support now or in the near-future for this feature (like I need to do some private demo tomorrow without any commitment), would you still choose the same? (elixir implementation in `web` and custom protocol).

At the end, as you repeat quite often, this environment is evolving really quickly and it is actually really overwhelming.
The thing that I most fear is that we choose something that we regret sooner than later and could have long-lasting effects in our delivery.
In that sense, I at least agree that a elixir based house-made code looks the simpler to deliver, as we don't need to package more things (which is the most frightening thing).

It is not a question for you, most probably more for @abravosuse , but I wonder if we could have any option to make this feature some sort of non-supported `tech-preview` in the 1st release. Like if we decided to use a 3rd party container, we could just ship it in `PackageHub` or so without real long-term support commitment.

gagandeepb (commented):
Thanks for this, Nelson! I left a few comments.

content/rfc/0004-ai-agent-integration.adoc
```diff
@@ -0,0 +1,295 @@
...
+* As a user, I want to configure my AI integration (e.g. provider and model selection, API key input) directly from the UI, so that I can easily set up and customize my AI assistant (Onboarding/Configuration)
+* As a user, I want an AI assistant chatbox easily accessible within the UI that can understand complex, multi-step requests, so that I can get help and information to solve my problems more efficiently (Conversation)
+
+There are also some more nuances that can be added to the above use cases:
+
+* the conversation should flow naturally without abrupt stops or "forgetting" previous parts
+* the system should handle large conversations efficiently to keep the chat going smoothly
+* the system should be able to handle complex requests by breaking them down and using the best tool or function for each part
+* the system should have access to relevant UI context to provide more accurate and helpful responses
+* the system should have access to relevant product-specific context (e.g. RAG) to provide more accurate and helpful responses
```
balanza, line 47: **suggestion:** ‏I would make the difference clear between the context coming from the domain knowledge, and the context coming from the current state of the application. They might have different requirements and lead to different architectures

content/rfc/0004-ai-agent-integration.adoc
```diff
@@ -0,0 +1,295 @@
...
+* the link:_detailed_design[premise]
+* the fact that we are considering not going AG-UI, yet
+* vendor lock-in/licensing risks (mainly with CopilotKit)
+* branding/watermark issues
+
+The proposed direction is to defer commitment to a specific AG-UI (or non-AG-UI) compliant component library.
+
+== Summary
+
+The proposed design for the AI assistant chatbox is to implement it as a feature within Trento Web, leveraging an Elixir-based implementation of the LangChain framework for the agentic capabilities, using WebSockets for real-time communication between the UI and the agent, and deferring the adoption of AG-UI or any specific UI component library to a later stage.
```
balanza: I see no evil here. 

All the taken decisions are reasonable and, more importantly, they look easy to reverse in case we find better options along the way. Not much more to discuss here from my side, we can discuss more details on the PoC, I think. 

I'm more concerned about the deferred decisions (RAG yes/no, UI interaction, MCP). I wonder if they can disrupt the overall architecture once we address them.

content/rfc/0004-ai-agent-integration.adoc
```diff
@@ -0,0 +1,295 @@
...
+
+* native support in the tech stack with Socket/Channels
+* already used for other features, meaning that we can leverage existing infrastructure
+
+==== Protocol
+
+When it comes to the protocol, that is the semantics and structure of the messages exchanged between the UI and the AI Agent, there are two main options:
+
+* *AG-UI*: adopt an existing protocol for agent-user interaction, such as the Agent-User Interaction Protocol https://docs.ag-ui.com/introduction[AG-UI]
+* *Custom Protocol*: implement a protocol tailored to our specific use case and requirements
```
balanza, line 311: My 2 cents here.

Precisely because the topic is relatively new in the industry, and best practices are not yet consolidated, I would follow one protocol for the sake of keeping the pace with the community. I would just assess how impactful/heavy it is.

(I have no particular sympathy for this one; it can be any other).

content/rfc/0004-ai-agent-integration.adoc
```diff
@@ -0,0 +1,295 @@
...
+|===
+|Feature Name |AI assistant chatbox
+|Start Date |Jan 1st, 2026
+|Category |AI, Architecture
+|PR |https://github.com/trento-project/docs/pull/164[#164]
+|===
+
+== Summary
+
+Build an **AI assistant widget**, fully integrated within Trento product experience.
```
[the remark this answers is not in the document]
nelsonkopliku: what does that refer to, more specifically?

content/rfc/0004-ai-agent-integration.adoc
```diff
@@ -0,0 +1,295 @@
+== 4. AI Assistant Chatbox
+
+[width="100%",cols="<18%,<82%",]
+|===
+|Feature Name |AI assistant chatbox
```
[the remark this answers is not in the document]
nelsonkopliku: I changed here 8e2b20dfd7fc575acbf78b4d4475fb13a402d558 and highlighted the native nature in the "proposed path forward"

content/rfc/0004-ai-agent-integration.adoc
nelsonkopliku: I am partially understanding.

The main context is the state of the application (accessible via APIs and/or UI state), and that state is derived from domain knowledge.

So, in a sense, by having access to the state you already got access to the domain knowledge that led to that state.

What am I not considering? Or what would you suggest adding/rephrasing?

content/rfc/0004-ai-agent-integration.adoc
nelsonkopliku: Can you help me better understand the concerns?

**RAG yes/no**
it is in the plans, just not for the very first iteration.
There has been some exploration around this specific library in the elixir ecosystem https://github.com/georgeguimaraes/arcana.

The main open point around RAG is whether we demand the user to do the ingestion on their servers, or if we ship pre-crunched data that we might do in our building pipelines. Or a combination of both (SUSE provided vector database + the ability for the customer to add up their own content)

Another RFC will very likely follow for this.

**UI interaction/MCP**: can you elaborate the concerns?

content/rfc/0004-ai-agent-integration.adoc
nelsonkopliku: Yes, I am not against following a standard, I actually aim to that, and I have no particular preference over any.

In the context of AG-UI it is worth considering the following

> * There are means of [translating existing protocols to AG-UI](https://docs.ag-ui.com/quickstart/middleware)
> * There is work in progress in the AG-UI ecosystem to support Elixir [ag-ui/pull/1046](https://github.com/ag-ui-protocol/ag-ui/pull/1046) and [ag-ui/pull/1293](https://github.com/ag-ui-protocol/ag-ui/pull/1293)
> * We can implement it ourselves also partially using the parts of the protocol that are relevant to us
>

I don't exclude we might actually implement it, however I am wondering about the following:

- if we implement it right away, when the elixir support comes we might need to rework in favor of the community provided library
- if we don't implement it right away, when the elixir support comes we might need to rework in favor of the community provided library
- implementing it right away, necessarily means adding overhead

Additionally
- it is true that if we go AG-UI protocol from the beginning (even though with our own implementation) when a community supported elixir library/sdk comes, we would ideally (because things never go as planned) need to adapt only the backend, because the UI side would already be AG-UI compliant
- with a totally custom protocol, we'd need to adapt backend _and_ UI

That said, I'd find it reasonable keeping the amount of rework in the future limited, and thus implement the standard (hoping it remains so 🙈) protocol ourselves (even partially) right away.

However I don't have visibility, yet, on the impact/delay it might have on releasing the first version in next trento release.

We need to make the decision about our short and mid-term goals, the debt we want to take and where to re-scope.

Bottom line, I don't exclude any option. We need to look closer into that and make the call.

content/rfc/0004-ai-agent-integration.adoc
balanza: > The main context is the state of the application (accessible via APIs and/or UI state),

I mean, there will be a moment in which the model queries the Trento API besides what is actually rendered on screen. Anyway, it's just phrasing, not a big deal. As long as the concept is covered, I'm fine 👍🏼

content/rfc/0004-ai-agent-integration.adoc
balanza: I mean, many decisions are deferred, and it's not clear to me whether they have been framed already or whether they might have an impact on the decisions actually taken in this RFC. 

Given that I wasn't part of the discussion that led to this RFC, there might be some understanding in the team that I'm failing to grasp. Looking forward to the next iterations to contribute, though,

content/rfc/0004-ai-agent-integration.adoc
balanza: I think in the topic of AI agents, we cannot prevent/limit future rework as the scenarios are going to change unpredictably. So either way, we might need to revisit some parts in the near future. And what is claimed to be a _standard_ today can be legacy in 6 months 🤷🏼 

My 2 cents were more on the R&D investment: on the UI, we might study someone else's work rather than tailor our own solutions, so as to focus on things more related to our domain, like how to have the model understand the cluster topology.

Anyway, no big deal either way

content/rfc/0004-ai-agent-integration.adoc
```diff
@@ -0,0 +1,297 @@
...
+JS:
+
+* https://github.com/trento-project/liz/tree/TRNT-4140-1
+
+Elixir:
+
+* https://github.com/trento-project/web/tree/liz-testing-langchain-elixir
+* https://github.com/trento-project/web/tree/ai-native-poc
+
+Golang: no reference available, but it has been attempted.
```
antgamdia: It is available in this branch of the mcp-server repo: 
https://github.com/trento-project/mcp-server/tree/TRNT-4140-liz-alt

content/rfc/0004-ai-agent-integration.adoc
nelsonkopliku: Alright, this PR https://github.com/trento-project/web/pull/4186 aims to shed some lights around this topic.

If we like the direction I will change the RFC accordingly

content/rfc/0004-ai-agent-integration.adoc
nelsonkopliku: > Given that I wasn't part of the discussion that led to this RFC, there might be some understanding in the team that I'm failing to grasp
>
This RFC aims exactly to make room for a broader involvement and agreement. Thanks for the engagement and feedback!

Let me summarize:

- **UI interaction** should be covered here https://github.com/trento-project/web/pull/4186
- **RAG yes/no** I would reiterate the previous comment https://github.com/trento-project/docs/pull/164#discussion_r3032167054
- **MCP** is an interesting point, I will add a broader comment for this https://github.com/trento-project/docs/pull/164#issuecomment-4251257423 and related commit https://github.com/trento-project/docs/commit/0ce4a333189dfbfe1530cddd9e52c5b501862628

content/rfc/0004-ai-agent-integration.adoc
nelsonkopliku: Thanks! Added here 8a67b1dda5aeb597fe4c7814cf1fdf147fb32b41

content/rfc/0004-ai-agent-integration.adoc
```diff
@@ -289,6 +293,79 @@ Considering:
...
+
+PROs:
+
+* one less component to install and still have access to the same AI capabilities
+* opens the possiblity to expose internal functions as tools for the AI agent without necessarily exposing them as APIs, if not needed
+* less latency/networking overhead, as the calls to the tools would be mainly internal function calls instead of network calls to the MCP server
+* maximised authnz reuse within the same conversation context
+
+CONs:
+* exposing internal features/functions as tools for the AI agent might lead to repetition of code that is already wired up in how our controllers work, or that needs to be wired up both as a controller action and as an AI agent tool
```
nelsonkopliku, line 396: This is what I am referring to https://github.com/trento-project/web/pull/4171/changes#diff-ca5ffdf1282611fca6aea00cea87d9d600dcfbb794d449a621217a9c223b1b7bR15

There might be different ways to do the same mapping and reduce repetition.

nelsonkopliku:
@trento-project/developers the last commit 0ce4a333189dfbfe1530cddd9e52c5b501862628 addresses the Tools Integration/MCP topic. Please take a look.

nelsonkopliku:
In the last commit I updated the sections about the protocol and the UI implementation based on the findings and results of this family of PRs:

- https://github.com/trento-project/web/pull/4245
- https://github.com/trento-project/web/pull/4246
- https://github.com/trento-project/web/pull/4251
- https://github.com/trento-project/web/pull/4254

antgamdia (approved):

EMaksy:
Hey @nelsonkopliku lgtm, just don't forget to add the rfc in the navigation https://github.com/trento-project/docs/blob/main/trento-docs-site/modules/developer/nav_developer.adoc?plain=1#L40 and if the content is fine for the others we can merge :)

content/rfc/0004-ai-agent-integration.adoc
```diff
@@ -0,0 +1,434 @@
...
+
+* clear separation of concerns between the agentic logic and the web application logic
+* an example for the future about other possibly extractable pieces of code 
+
+CONs:
+
+* possibly higher level of friction in the development and evolution of the agentic runtime
+* we might not get the correct library boundaries from the start, meaning that we might end up with some parts in web that should be in the library, leading to friction and overhead in the development (like: is the Channel implementation part of the library or of web? is the AG-UI protocol implementation part of the library or of web? etc...), however when we realize this we can always move the code in the right place
+
+Given the above, the proposed solution is TBD...
```
nelsonkopliku: My personal take on this:

I appreciate the library approach for its cleanness, however what concerns me is the natural and inevitable friction in this early stage of implementation where web and the agentic runtime \*are\* necessarily coupled from a functional perspective (does not mean the code should be coupled though). 

In fact I recently found myself modifying the library in `./deps` for a shorter feedback loop.

I deemed necessary sharing the options for transparency and for a broader agreement.

nelsonkopliku:
@trento-project/developers the last commit 8695c05a237597c1071d332c06bc45aaf115505b adds a section about integration options of AI Agentic features into trento web.

content/rfc/0004-ai-agent-integration.adoc
antgamdia: +1, adding more projects/repos/dependencies is just increasing the complexity with no clear benefit to me at this very moment. At least, without a clear use case: is this agentic runtime also expose the A2A protocol so that it can be consumed by other agents? Or does it always require an interaction with web (auth, etc)?
We are even re-implementing the MCP tools inside web... so doing so and yet adding a new piece of code... seems a bit inconsistent to me :S

nelsonkopliku: @balanza bringing your comment at https://github.com/trento-project/docs/commit/8695c05a237597c1071d332c06bc45aaf115505b#commitcomment-185277146 here as it might be simpler to follow

> I am not sure I got the point. Is the comparison between having Sagents as a dependency of Trento web vs having Sagents as a dependency of a library that is a dependency of Trento web? I don't see the value of the indirection; thus, what is the benefit?
> 
> BTW, I'm not working directly on the topic, and I may miss the nuances. What is the opinion of the actual team working on it? @nelsonkopliku @antgamdia @gagandeepb 
>

The thing is exactly what you mentioned: bringing Sagents features directly into Trento web or wrapping Sagents into a lib that is integrated in Trento web.

Maybe I should have been more explicit [here](https://github.com/trento-project/docs/pull/164#discussion_r3239963415), however based on the outlined pros and cons of the two approaches I would personally find it safer and acceptable at this stage starting from integrating Sagents directly in Trento web.

arbulu89: If the decision was up to me, I would've add the code in `web` directly.
The single reason of `cleanness` for me is not enough to have a new repository, because at the end, this `web` is the unique user of it, and it is written in the same language.
Having additional repos add operational work that adds-up.
If we had more valid reasons than "code structure" it would be a different thing.

Now, I have not even really looked into the new repository, so my comment here is just really high-level opinion.

nelsonkopliku: Alright, thanks for the feedback. I will move into web.

chargio:
Consider the idea of making this an application inside an umbrella:
1. Same configuration and dependencies, but independent applications.
2. Independent applications can be tested in isolation, without the need to test them together, although they can also be tested together.
3. Each application can have its own supervisor tree, so if they fail the rest of the application does not.
4. You can define at runtime if you want the application to run or not, using configuration

content/rfc/0004-ai-agent-integration.adoc
nelsonkopliku: https://github.com/trento-project/web/pull/4320

PR #166, opened 2026-04-07
[TRNT-4330] Add Dependabot configuration

antgamdia:
This PR adds the configuration for Dependabot, so that we can get automated updates when the dependencies change. This configuration also includes a cooldown period, to align with the configuration used in other repositories.

Related # TRNT-4330


EMaksy (approved):
Thanks for this, well  i guess a lot of dependabot issues will be open up with this one. 
I think  i will take care of it overtime, but it is a good step forward 👍

PR #188, opened 2026-04-22
RFC: predicates for host exclusion in checks

balanza:
[Visual](https://github.com/trento-project/docs/blob/check-exclusion-predicate/content/rfc/0004-check-exclusion-predicates.adoc)

### DIFF
--- a/content/rfc/0004-check-exclusion-predicates.adoc
+++ b/content/rfc/0004-check-exclusion-predicates.adoc
@@ -1,0 +1,158 @@
+== 4. Host-level Exclusion Predicates for Check Executions
+
+[width="100%",cols="<18%,<82%",]
+|===
+|Feature Name | Host-level Exclusion Predicates for Check Executions
+|Start Date | Apr 22nd, 2026
+|Category | Architecture, Checks
+|PR | TBD
+|===
+
+== Summary
+
+Let a check declare an optional `exclude:` Rhai predicate; Wanda evaluates it per target using host metadata supplied by Trento Web, and drops excluded hosts before dispatching the gathering request.
+
+== Motivation
+
+Today a check execution targets every host the user selected in the Web UI. There is no way for a check author to declare "this check does not apply to host X" based on properties of the host itself.
+
+We want check authors to express, inside the check definition, a predicate that decides per-host whether the check should run at all, evaluated against data the Web UI already knows about each host.
+
+=== Use Cases outline
+
+* As a check author, I want to declare that a check does not apply to hosts with certain properties (e.g. provider, OS family, cluster role), so that I don't need to maintain near-duplicate checks.
+* As a user running an execution, I want excluded hosts to be clearly surfaced in the result, distinct from failed or skipped ones, so that I can tell why a host was not evaluated.
+* As a catalog maintainer, I want applicability logic to live next to the check it belongs to, instead of being scattered across Web, Wanda and catalog curation.
+
+== Detailed Design
+
+=== Overview
+
+* Trento Web attaches a small, structured "host data" payload to each target in the execution request.
+* The execution-request contract is extended so each target carries its host data alongside `agent_id` and `checks`.
+* A check definition gains an optional top-level `exclude:` field, containing a Rhai expression evaluated against the host data.
+* Wanda evaluates the `exclude` predicate per (host, check) pair and filters out excluded pairs before dispatching `FactsGatheringRequested` to the agents.
+* Excluded pairs are reported in the execution result with a clear "excluded by policy" status, distinct from failed or skipped, so the UI can render it.
+
+=== Component changes
+
+==== Trento Web
+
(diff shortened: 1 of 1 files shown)


content/rfc/0004-check-exclusion-predicates.adoc
```diff
@@ -0,0 +1,158 @@
...
+|PR | TBD
+|===
+
+== Summary
+
+Let a check declare an optional `exclude:` Rhai predicate; Wanda evaluates it per target using host metadata supplied by Trento Web, and drops excluded hosts before dispatching the gathering request.
+
+== Motivation
+
+Today a check execution targets every host the user selected in the Web UI. There is no way for a check author to declare "this check does not apply to host X" based on properties of the host itself.
```
nelsonkopliku: **nit:**

> a check execution targets every host the user selected in the Web UI
>

the user does not actually select hosts for checks execution in the web UI.

The user decides to run checks agains a given target, and that target can be either:
- a specific host (in this case we might end in the _user selects a host in the UI_ scenario)
- a cluster (at this time all the hosts part of a cluster are targeted)

content/rfc/0004-check-exclusion-predicates.adoc
```diff
@@ -0,0 +1,158 @@
...
+    repeated string checks = 2;
+
+    // Free-form host metadata known to Trento Web at request time.
+    // Consumed by Wanda to evaluate each check's `exclude` predicate;
+    // never forwarded to agents.
+    //
+    // Keys are opaque strings agreed between Trento Web (producer) and
+    // check authors (consumers); the contract on which keys exist and
+    // what they mean lives outside the proto file.
+    map<string, google.protobuf.Value> host_data = 3;
```
nelsonkopliku: **thought:** what about stripping the `host` reference from this new piece of information, and keep the `target` wording.

Instead of `host_data` we could have `data`,`target_data`,`metadata`, `target_metadata`

content/rfc/0004-check-exclusion-predicates.adoc
```diff
@@ -0,0 +1,158 @@
...
+* `true` => exclude this host from this check.
+* Absent field => never exclude (current behaviour).
+
+.Example: exclude majority maker nodes from a cluster check
+[source,yaml]
+----
+id: "ABC123"
+name: Check that does not apply to majority maker nodes
+exclude: |
+  host["is_majority_maker"] == true
```
nelsonkopliku: **thought:** we could consider consistently advertise dot access like `host.is_majority_maker`.

IIRC accessing a non existent entry would return `()` anyway, right?

content/rfc/0004-check-exclusion-predicates.adoc
```diff
@@ -0,0 +1,158 @@
...
+
+1. Evaluates each check's `exclude` predicate against each target's host data.
+2. Filters the (host, check) pairs accordingly.
+3. Reports excluded pairs in the execution result with a clear "excluded by policy" status, distinct from failed / skipped, so the UI can render it.
+
+The predicate runs in the same Rhai sandbox Wanda already uses for checks; no new sandboxing is introduced.
+
+== Non-goals
+
+* Changing how `when:` conditions work inside expectations.
```
nelsonkopliku: **nit:** AFAIR there is no `when` inside expectations.

`when` can be either:
- at root level driving selectability/availability of the check for the execution based on the context
- at condition level defining the condition expression

nelsonkopliku (approved):
Looks good to me, Emanuele.

Just a tiny minor comments.

content/rfc/0004-check-exclusion-predicates.adoc
balanza: AFAIK given it's a map should be accessed by index, but I have to test it.

content/rfc/0004-check-exclusion-predicates.adoc
balanza: That's what I meant, but it was poor phrasing. Fixed

arbulu89 (commented):
Looks good @balanza ,
I have only one question. It is about the `excluded_by_policy` return value.
You say there is not need to do UI changes from the frontend, but at the same time, we are adding this new return option.
And even see some changes proposed in `web` repository to updated this.
Does it makes sense to update that paragraph where you say `No UI change is required for v1: the data is already available server-side; we just need to forward it.` to adjust to the new reality?

content/rfc/0004-check-exclusion-predicates.adoc
```diff
@@ -0,0 +1,158 @@
+== 4. Host-level Exclusion Predicates for Check Executions
+
+[width="100%",cols="<18%,<82%",]
+|===
+|Feature Name | Host-level Exclusion Predicates for Check Executions
+|Start Date | Apr 22nd, 2026
+|Category | Architecture, Checks
+|PR | TBD
```
arbulu89: maybe we can link the the 3 PRs we have produced (web, wanda and wanda docs)

balanza: Done in https://github.com/trento-project/docs/pull/188/changes/1a1b24f96d9088614c9c4ab9f2594928a075b119

arbulu89 (commented):
Hey,
Last nitpick comment, I promise 🙈 
Can you replace the `excluded by policy` usages by simply `excluded`?
They are some few, mixed with and without underscores

arbulu89 (approved):
💚 Thank you @balanza

PR #191, opened 2026-05-06
Add resource refresh operation and rename cluster start/stop operation name

arbulu89:
Three things:
- Add resource refresh operation
- Rename cluster start/stop operation name to use the new name
- Fix resource maintenance operation pre-condition text


abravosuse (approved):
LGTM!

PR #198, opened 2026-05-12
Ansible support

EMaksy:
# Description
The server, agent, and site playbooks now install supportutils + supportutils-plugin-trento by default on SLES for SAP 15 SP5+, so supportconfig archives include Trento diagnostics. 
This change brings the docs in line with that behavior.
## Changes
- Document the new `supportutils` Ansible role introduced in trento-project/ansible#106
- Add a `Diagnostics tooling` entry under **Components** describing the role
- Add a new **Support tooling** subsection covering configuration variables


skrech (approved):
Thank you, @EMaksy !

PR #199, opened 2026-05-13
Document operation failed new fields for activity log

arbulu89:
Document operation failed new fields for activity log


abravosuse (approved):
LGTM!

PR #201, opened 2026-05-20
Visualization of versions of different components in About view

abravosuse:
The About view now shows the versions of all components of Trento Server, not only web.


arbulu89 (approved):
Thank you!

PR #202, opened 2026-05-20
Filesystem and swap utilization dashboards

abravosuse:
The host details view now includes filesystem and swap utilization dashboards in addition to the CPU and memory utilization ones.


arbulu89 (approved):
I guess the images goes away and it is not replaced, right?

abravosuse:
> I guess the images goes away and it is not replaced, right?

Yes. The current image becomes obsolete. And replacing it with a new one simply adds maintenance overhead in the future. Actually, now that we have a new person in the Docs Team, I want to discuss with her the possibility of removing -if not all- at least the ones subject to possible future changes.

Thanks for the review!

PR #203, opened 2026-05-20
trento_supportconfig_plugin_generates_scenario_dump

abravosuse:
The Trento supportconfig plugin generates a scenario dump when executed in systemd scenarios.


arbulu89 (approved):
good!

PR #205, opened 2026-05-22
Updates on helm deployments procedures regarding the write endopoint

abravosuse:
See https://jira.suse.com/browse/TRNT-4424.

- Section https://documentation.suse.com/sles-sap/trento/html/SLES-SAP-trento/id-installation.html#sec-trento-k8s-deployment:
-- Add the following flags to all the helm installation commands included in the section: --set prometheus.enabled=false 
-- Add note explaining that this delivers an installation of Trento Server with no prometheus
- Section https://documentation.suse.com/sles-sap/trento/html/SLES-SAP-trento/id-prometheus-integration.html#id-kubernetes-deployment:
-- Re-write the section explaining that when prometheus is enabled in a K8s installation, the write end point is enabled by default and therefore it requires specifying what type of authentication we want to use and, in case of basic authentication, also the username and password. The refer the user to the section where it is explained how to the "Enabling Prometheus remote write for Grafana Alloy hosts" for further details.
- Section https://documentation.suse.com/sles-sap/trento/html/SLES-SAP-trento/id-update.html#sec-trento-updating-trentoserver:
-- explain that when updating to 3.1 in a helm installation, prometheus authentication type must be specified and, in the case of basic, username and password must be provided.

### DIFF
--- a/trento/adoc/trento-integration-prometheus.adoc
+++ b/trento/adoc/trento-integration-prometheus.adoc
@@ -1,6 +1,7 @@
 include::product-attributes.adoc[]
 
 :revdate: 2026-04-30
+[[sec-prometheus-integration]]
 == {prometheus} integration
 
 {prometheus} Server is a {trserver} component responsible for retrieving host metrics from the agent hosts and serving them to the web component. The web component renders some of these metrics as graphical charts in the details view of the registered hosts. Particularly: memory, CPU, filesystem and swap utilization.
@@ -38,9 +39,9 @@ See <<configuring-alloy>> for detailed setup instructions.
 
 === {kube} deployment
 
-When using the {helm} chart to deploy {trserver} on a {kube} cluster, an image of {prometheus} Server is deployed automatically. No further actions are required by the user, other than ensuring that it can reach the `node-exporter` in the agent hosts.
+To enable {prometheus} alongside {trserver} in a {k8s} cluster, see <<enabling-remote-write>>.
 
-In a {kube} cluster with multiple nodes, the user can select on which node to deploy the {prometheus} Server by adding the following flag to the {helm} installation command:
+In a {kube} cluster with multiple nodes, the user can select on which node to deploy {prometheus} Server by adding the following flag to the {helm} installation command:
 
 [source,bash]
 ----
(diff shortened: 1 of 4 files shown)


trento/adoc/trento-integration-prometheus.adoc
```diff
@@ -38,9 +39,9 @@ See <<configuring-alloy>> for detailed setup instructions.
 
 === {kube} deployment
 
-When using the {helm} chart to deploy {trserver} on a {kube} cluster, an image of {prometheus} Server is deployed automatically. No further actions are required by the user, other than ensuring that it can reach the `node-exporter` in the agent hosts.
+To enable {prometheus} alongside {trserver} in a {k8s} cluster, see <<enabling-remote-write>>.
```
antgamdia, line 42: question: I assume the change from `{k8s}` to `{kube}` does not affect, no? Is it still rendering "Kubernetes", no?

trento/adoc/trento-kubernetes-install.adoc
```diff
@@ -180,10 +187,10 @@ HELM_EXPERIMENTAL_OCI=1 helm upgrade \
    --install trento-server oci://registry.suse.com/trento/trento-server \
    --set global.trentoWeb.origin=TRENTO_SERVER_HOSTNAME \
    --set trento-web.adminUser.password=ADMIN_PASSWORD \
-   --set prometheus.server.nodeSelector.LABEL=VALUE \
    --set postgresql.primary.nodeSelector.LABEL=VALUE \
    --set trento-web.nodeSelector.LABEL=VALUE \
-   --set trento-runner.nodeSelector.LABEL=VALUE
+   --set trento-runner.nodeSelector.LABEL=VALUE \
```
antgamdia: note [out of the scope]: I'm not 100% sure, but I'd say `trento-runner` is not used in the chart.

trento/adoc/trento-kubernetes-install.adoc
antgamdia: note [out of the scope]: it's been quite a while since helm 3.8 is out, we even have helm 4 today . At some point, we're gonna need to remove the experimental flag 😅

antgamdia (approved):
Thanks!

trento/adoc/trento-integration-prometheus.adoc
abravosuse: Yes. Both variables have the same value defined in the generic_attributes.adoc file.

trento/adoc/trento-kubernetes-install.adoc
abravosuse: Good catch! I will push a commit to remove it!

trento/adoc/trento-kubernetes-install.adoc
abravosuse: I agree with you. I will push a commit removing the experimental flag.

abravosuse (commented):
Thank you @antgamdia . I will push a couple of commits to address your remarks/suggestions.

PR #211, opened 2026-06-15
Fix documentation links

eromanova97:
### Description
Fixing in-line links to not render as clickable. 
For reference, you can check the html build, chapters `4 Installation` and `8 MCP Integration`
 For example:

Previously:
image

Now:
image

### Reasoning
Rendering dummy links, such as `example.com`, unclickable is beneficial because it prevents automated documentation link checkers from attempting to validate a generic placeholder.

This change will go to `latest` as well.


antgamdia (approved):
Thanks!

PR #210, opened 2026-06-11
Update Prometheus installation instructions on SLES for SAP 16.0 (#209)

eromanova97:
Cherry-pick of https://github.com/trento-project/docs/pull/209

PR #214, opened 2026-06-17
Add information about updating CA certificates on Trento Server host

eromanova97:
### Description

Add missing information about updating CA certificates.
Fixes [TRNT-4398](https://jira.suse.com/browse/TRNT-4398)

#### Branches:
`main`, `latest`

**NOTE TO MYSELF BEFORE MERGING:** need to update revision history for this and for https://github.com/trento-project/docs/pull/210

### DIFF
--- a/trento/adoc/trento-systemd-install.adoc
+++ b/trento/adoc/trento-systemd-install.adoc
@@ -2,7 +2,7 @@ include::product-attributes.adoc[]
 
 [[sec-systemd-deployment]]
 ==== systemd deployment
-:revdate: 2025-06-11
+:revdate: 2026-06-19
 
 A systemd-based installation of the {trserver} using RPM packages can be performed manually on the latest supported versions of {sles4sap}, from 15 SP4 up to 16. For installations on service packs other than the current one, make sure to update the repository URL as described in the relevant notes throughout this guide.
 
--- a/trento/adoc/trento-systemd-install.adoc
+++ b/trento/adoc/trento-systemd-install.adoc
@@ -509,6 +509,11 @@ mkdir -p /etc/nginx/ssl/certs/
 ----
 cp trento.crt /etc/nginx/ssl/certs/trento.crt
 ----
++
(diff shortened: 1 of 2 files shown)


trento/adoc/trento-systemd-install.adoc
```diff
@@ -509,6 +509,11 @@ mkdir -p /etc/nginx/ssl/certs/
 ----
 cp trento.crt /etc/nginx/ssl/certs/trento.crt
 ----
++
+[IMPORTANT]
+====
+Update the CA certificates on the {trserver} host to ensure the self-signed certificate is trusted.
```
eromanova97: Question for the reviewer:
The steps:
```
cp trento.crt /etc/pki/trust/anchors/
update-ca-certificates
```

Are there other methods? If this is the main method we want users to follow, then I would, instead of a note, add a proper step with commands to do this.

eromanova97:
Screenshot of the rendered step for reference: 
image

trento/adoc/trento-systemd-install.adoc
abravosuse: Good point @eromanova97 . A step would be better than a note. The only doubt I have is that we might have to convert the crt format into pem one before placing it under `/etc/pki/trust/anchors/`. At least that is what we tell customers to do on the Agent hosts. @nelsonkopliku ?

eromanova97:
Here is the screenshot of the command version:
image

trento/adoc/trento-systemd-install.adoc
nelsonkopliku: > A step would be better than a note
>
agreed

> we might have to convert the crt format into pem one before placing it under /etc/pki/trust/anchors/
>
technically speaking there might not be a strict need but since 
> that is what we tell customers to do on the Agent hosts
>
then I would keep it consistent

nelsonkopliku (approved):
Thank you!

PR #229, opened 2026-07-09
Add information about trentoWeb.origin parameter

eromanova97:
### Description

Based on the thread in Slack, we could enhance the docs by adding some additional explanation for the  trentoWeb.origin parameter; I took the opportunity to add some additional small things.

Fixes https://jira.suse.com/browse/TRNT-4476
### Additional information

I have added it in two places, since both are separate user cases. Plus I added small extra explanation line to the "selected node" case.

Previews: 
[kubernetes installation - SLES-SAP-trento_en.pdf](https://github.com/user-attachments/files/29888153/kubernetes.installation.-.SLES-SAP-trento_en.pdf)

[RPM installation - SLES-SAP-trento_en.pdf](https://github.com/user-attachments/files/29888156/RPM.installation.-.SLES-SAP-trento_en.pdf)

### DIFF
--- a/trento/adoc/trento-kubernetes-install.adoc
+++ b/trento/adoc/trento-kubernetes-install.adoc
@@ -2,7 +2,7 @@ include::product-attributes.adoc[]
 
 [[sec-trento-k8s-deployment]]
 ==== {k8s} deployment
-:revdate: 2026-07-09
+:revdate: 2026-07-10
 
 include::partials/kubernetes-placeholders.adoc[]
 
@@ -73,7 +73,7 @@ helm upgrade \
 Where:
 
 `global.trentoWeb.origin`::::
-Uses the provided hostname to configure internal functionalities such as the interactive API portal, websockets, or certificate generation.
+Uses the provided hostname (for example, `trento.example.com`) to configure internal functionalities such as the interactive API portal, websockets, or certificate generation.
 
 `trento-web.adminUser.password`::::
 Sets the initial password for the default admin account to access the {trento} console.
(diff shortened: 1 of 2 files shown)


trento/adoc/trento-kubernetes-install.adoc
```diff
@@ -137,6 +149,17 @@ helm upgrade \
 ----
 ====
 +
+Where:
+
+`global.trentoWeb.origin`::::
+Uses the provided hostname to configure internal functionalities such as the interactive API portal, websockets, or certificate generation.
```
antgamdia: I'll have to double-check, but I'm not fully sure whether this param requires a scheme to also be present. Like in `https://TRENTO_SERVER_HOSTNAME`

Anyways, maybe we can have an example value, like `trento.example.com` so that users can clearly know what the hostname, etc., is

eromanova97: Great catch! especially since we have: `http://TRENTO_SERVER_HOSTNAME` after, it would seem like it needs to be passed without the scheme. Let me know what you find out when you double-check, thank you!

antgamdia: It seems trentoWeb.origin is just the hostname, not the whole URL.

References:
- https://github.com/trento-project/web/blob/a60bbd9e9e2a2d4fecf7220f6dd15e1bfa47b66a/lib/trento/release.ex#L149
- https://github.com/trento-project/web/blob/a60bbd9e9e2a2d4fecf7220f6dd15e1bfa47b66a/.github/workflows/ci.yaml#L501

I got it confused by `OAS_SERVER_URL`, which has the scheme and port 😅 

https://github.com/trento-project/web/blob/main/guides/OpenAPI/openapi.adoc

I'd still add an example, to make it easier for users to understand.

antgamdia (approved):

antgamdia:
By the way, I see the change is affecting the k8s installation; however, I think we still to mirror it in the usual RPM installation (the env variable should be `TRENTO_WEB_ORIGIN`)

eromanova97:
I see! I will update it as part of this PR then :+1:

PR #232, opened 2026-07-13
RFC for internal MCP server in substitution of the current go version

chargio:
### Description

RFC to describe the actions required to implement the MCP server as a web server instead of an external go application.

Fixes #<issue>
Reduces complexity of the implementation and test of the MCP server, reducing duplicated code.

### Documentation changes

Yes


content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+It is possible to simplify the architecture and installation of Trento by substituting the MCP server with an Elixir MCP server that reuses the tools developed for Liz.
+
+==== Use Cases outline
+
+* As a user, I want to be able to publish capabilities of Trento using the MCP protocol, so that it can be used by an external agent.
+* As a user, I want to have a similar experience when I am using the internal AI agent (Liz) and when I am using the external MCP agent, so that I can use Trento in a consistent way.
+* As a user, I want to be able to enable/disable the MCP server
+* As a user, I want to be able to configure the MCP server.
+* As a user, I want to use MCPHost to connect to the MCP server in SLES 16.
```
antgamdia:
```suggestion
* As a user, I want to use MCPHost/Kit to connect to the MCP server in SLES 16.
```

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+Considering the outlined link:#use-cases-outline[use cases] we need to:
+
+* define the capabilities already present in the MCP server and how they are used by the external agent.
+* define the changes required to migrate to an internal MCP server, and what would be the tradeoffs.
+* define a migration path to move from the current MCP server to the new internal one.
+
+THe current MCP server is a Go application that implements the MCP protocol, and it is used by the external agent to communicate with Trento. Installing the MCP server requires indepedent configuration and connection to an existing installation, introducing a bigger attack surface and configuration problems. The Trento MCP server requires:
+- A running Trento server (version 3.x or later).
+- An MCP-Compatible AI assitant
```
antgamdia:
```suggestion
- An MCP-Compatible AI assistant (MCP client)
```

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+== Detailed design
+
+Considering the outlined link:#use-cases-outline[use cases] we need to:
+
+* define the capabilities already present in the MCP server and how they are used by the external agent.
+* define the changes required to migrate to an internal MCP server, and what would be the tradeoffs.
+* define a migration path to move from the current MCP server to the new internal one.
+
+THe current MCP server is a Go application that implements the MCP protocol, and it is used by the external agent to communicate with Trento. Installing the MCP server requires indepedent configuration and connection to an existing installation, introducing a bigger attack surface and configuration problems. The Trento MCP server requires:
```
antgamdia:
```suggestion
The current MCP server is a Go application that implements the MCP protocol, and it is used by the external agent to communicate with Trento. Installing the MCP server requires independent configuration and connection to an existing installation, introducing a bigger attack surface and configuration problems. The Trento MCP server requires:
```

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+- A running Trento server (version 3.x or later).
+- An MCP-Compatible AI assitant
+- A Trento Personal Access Token (PAT), generated from your Trento user profile for secure API access.
+- Configuration in /etc/trento/mcp-server-trento
+
+The Trento MCP server today is a Go applcation, that acts as a bridge between the MCP clients and the Trento API, automatically generating MCP tools from the OpenAPI specification.
```
antgamdia:
```suggestion
The Trento MCP server today is a Go application that acts as a bridge between the MCP clients and the Trento API, automatically generating MCP tools from the OpenAPI specification.
```

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+The trento MCP server generates a list of tools based on the OpenAPI specification, using labels to define which tools are available for the MCP clients.
```
antgamdia: note: for that we actually use a 3rd-party dependency, https://pkg.go.dev/github.com/evcc-io/openapi-mcp, to which we usually contribute whenever we find bugs... but it is not widely used right now.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+You can run an MCP server directly using on the native Elixir MCP packages. The two main packages are:
+
+- `annubis_mcp`` (formely `hermes_mcp`). A comprehensive, production-ready SDK that supports STDIO (for local clients) and SSE/HTTP (for remote/web clients). It allows for the dynamic registration of tools. 
+- `emcp`. A minimalistic, lightweight MCP client/server library designed to be mounted directly as a Plug in your phoenix router.
+
+Wse also considered using Nodejs, building a tiny Node MCP server, using the official `@modelcontextprotocol/sdk`to build a server that fetches tool definitino for the Elixir backend and proxies execution requests back to an Elixir endpoint.
```
antgamdia:
```suggestion
We also considered using Node.js, building a tiny Node MCP server, using the official `@modelcontextprotocol/sdk`to build a server that fetches tool definitions for the Elixir backend and proxies execution requests back to an Elixir endpoint.
```

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+== Security
+
+The Go mcp-server acts as a multi-user proxy between the client (your IDE or agent) and the Elixir trento-web API. The job of the MCP is to fetch OpenAPI specs
+
+* *Token Delivery*: The MCP client (e.g., VS Code or Claude Desktop) is configured with an Authorization header containing the user's Trento Personal Access Token (PAT).
+* *Session Mapping*: When the client initializes the connection (via SSE/HTTP), the Go mcp-server intercepts the header, associates the PAT with the unique sessionID, and stores it.
```
antgamdia, line 58: note: the next protocol version is supposed to be totally stateless, so this is gonna become a problem if we want to support it.

antgamdia (commented):
Thanks for writing it up! Before jumping to major discussion, just some minor notes and typos.

Edit: copilot was faster 😅

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: Updating wording

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: I've added the link to the text.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: Currently we use a PAT to do the same, so the problem would be anyway present, and we will need to solve it anyway for Liz

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+4. You can't use command-line flags to configure the internal MCP server. The configuration is done via the Elixir config files or environment variables. Additional UX design should be done to include them as part of the command line for Trento itself.
+
+
+== Additional considerations
+- It would be optional to create a solution that can provide the MCP server as an application inside the BEAM (umbrella application). In that case, it would be possible to define via configuration if the MCP server is started or not, and it would be possible to run it standalone. In an umbrella application, tests for each application are independent. Umbrella applications would imply additional complexity, and it is not clear if it would be worth the effort. The current proposal is to have the MCP server as part of the Trento application, and it would be started automatically when Trento is started. An application inside the BEAM would be a nice to have, but it would require additional work to transition from a single application to a series of applications, in which case the MCP would be one of them.
```
gagandeepb: This RFC does not motivate the 'why' of the umbrella application solution. Why do you think the umbrella application solution (exlusively, perhaps ?) addresses some issues that the current trento-web application architecture does not or can not? Can you also please clarify if you want/are proposing a separate deploy-able artifact for the umbrella members?

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+-  The endpoint is currently defined as `trento-server/mcp-server_endpoint/mcp`. It would make sense to simplify the endpoint to be `trento-server/mcp`, as the endpoint is already defined in the Trento application, and it is not necessary to have a separate endpoint for the MCP server. The current proposal is to keep the endpoint as it is, but it would be nice to have a simpler endpoint or even double it and deprecate the old one.
```
gagandeepb: Can you please explain how is this consideration consistent with the desire/need to have a configurable`TRENTO_MCP_PORT`?

gagandeepb: Can you please explain further what would "even double it and deprecate the old one" mean in more concrete terms?

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+|===
+
+Not every option is superfluous, but the configuration is simpler and more aligned with the rest of Trento's configuration. The only options that remain are:
+
+[cols="3,1,6", options="header"]
+|===
+|Environment variable | default value | Detail
+| TRENTO_MCP_ENABLED | false | Whether the internal MCP server is enabled or not. If false, the internal MCP server will not be started.
+| TRENTO_MCP_PORT | 5000 | The port on which the MCP server listens for incoming connections with HTTP transport
+| TRENTO_MCP_LOG_LEVEL | info | The verbosity level of the MCP server logs. Possible values are: debug, info, warn, error
```
gagandeepb, line 102: Any thoughts/opinions on mapping MCP server logs to ActivityLogs?

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: That is not part of the RFC, it is something to consider for the future.
It would allow to test the MCP server indepednently of the rest. And start the application only when needed

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: Fair point

chargio: The current solution is puslished at the trento-server/mcp-server_endpoint/mcp endpoint in port 5000. With the new to decide if we want to have it in the same port or we want to have it in a different port (no longer needed as the pipeline is the same and not a different component in the same system requiring a separate port)

chargio: We can have both endpoints available in routes, the existing one and the simplified ones, pointing to the same application.
Then we can deprecate the old one and migrate in time to the new one.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: Can you ellaborate?

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+== Summary
+
+It is possible to simplify the architecture and installation of Trento by substituting the MCP server with an Elixir MCP server that reuses the tools developed for Liz.
+
+==== Use Cases outline
+
+* As a user, I want to be able to publish capabilities of Trento using the MCP protocol, so that it can be used by an external agent.
+* As a user, I want to have a similar experience when I am using the internal AI agent (Liz) and when I am using the external MCP agent, so that I can use Trento in a consistent way.
+* As a user, I want to be able to enable/disable the MCP server
```
nelsonkopliku, line 18: **note:** we need to define whether the MCP server is enabled/disabled at startup via env variables or via UI.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+* As a user, I want to be able to configure the MCP server.
```
nelsonkopliku, line 19: **question:** what kind of configuration does this entail?

Is it about configuring the endpoint/port where the MCP server listens to, if even necessary? Or is it about configuring which tools are exposed?

content/rfc/0005-migrate-to-internal-mcp-server.adoc
nelsonkopliku: **question:** are we talking about https://github.com/mark3labs/mcphost and https://github.com/mark3labs/kit? Or something else?

Let's add a reference in any case.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+* As a user, I want to use MCPHost to connect to the MCP server in SLES 16.
+* As a user, I want to use both streamable and sse as the transport protocols for the MCP server, so that I can use the one that is more suitable for my environment.
```
nelsonkopliku, line 21: **suggestion:** in latest MCP spec, SSE seems to be the `legacy` way https://modelcontextprotocol.io/specification/2025-11-25/basic/transports.

Let's agree which transport we should support.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+* As a developer, I want to reduce the maintenance needed for the alignment of tools used by Liz and by the MCP server, so that the customer gets the same capabilities and responses independently of how they use it.
```
nelsonkopliku, line 22: **note:** this is already the case currently. "MCP" tagged open api endpoints become automatically available tools to MCP server and Liz implementation.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+The current MCP server is a Go application that implements the MCP protocol, and it is used by the external agent to communicate with Trento. Installing the MCP server requires independent configuration and connection to an existing installation, introducing a bigger attack surface and configuration problems. The Trento MCP server requires:
+- A running Trento server (version 3.x or later).
+- An MCP-compatible AI assistant
+- A Trento Personal Access Token (PAT), generated from your Trento user profile for secure API access.
+- Configuration in /etc/trento/mcp-server-trento
+
+The Trento MCP server today is a Go application that acts as a bridge between the MCP clients and the Trento API, automatically generating MCP tools from the OpenAPI specification.
+
+The Trento MCP server generates a list of tools based on the OpenAPI specification, using labels to define which tools are available for the MCP clients.
```
nelsonkopliku: **suggestion:** rather than in the design of the under discussion solution, this section would better fit in the motivation section as it explains the current state and its drawbacks.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+|===
+| Setting | Default | Notes 
+|`config :trento, :mcp_server, enabled` | `false` | Gates the supervised server + makes `/mcp` inert when off 
+| `ENABLE_MCP_SERVER` (runtime env) | `false` | Sets the above in `config/runtime.exs` |
+|===
+
+
+=== Caveats
+
+1. The Go bridge that we were using before gives you the API for free, only taking into account the tags required to expose it. Using Anubis in Elixir is already automated. The same "MCP" tags used in the Trento server are used to configure the AI assistant's toolset and so that they continue being aligned automatically. The only new work is the authentication plug.
```
nelsonkopliku, line 126: **question:** what does "Using Anubis in Elixir is already automated" refer to?

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+- A Trento Personal Access Token (PAT), generated from your Trento user profile for secure API access.
+- Configuration in /etc/trento/mcp-server-trento
+
+The Trento MCP server today is a Go application that acts as a bridge between the MCP clients and the Trento API, automatically generating MCP tools from the OpenAPI specification.
+
+The Trento MCP server generates a list of tools based on the OpenAPI specification, using labels to define which tools are available for the MCP clients.
+
+You can run an MCP server directly using the native Elixir MCP packages. The two main packages are:
+
+- `anubis_mcp` (formerly `hermes_mcp`). A comprehensive, production-ready SDK that supports STDIO (for local clients) and SSE/HTTP (for remote/web clients). It allows for the dynamic registration of tools.
```
nelsonkopliku, line 48: **question:** what does "It allows for the dynamic registration of tools" effectively mean?

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+- `emcp`. A minimalistic, lightweight MCP client/server library designed to be mounted directly as a Plug in your Phoenix router.
+
+We also considered using Node.js, building a tiny Node MCP server that uses the official `@modelcontextprotocol/sdk` to fetch tool definitions from the Elixir backend and proxies execution requests back to an Elixir endpoint.
```
nelsonkopliku, line 51: **thought:** in the context of an "Internal MCP server" I think this option is basically swapping the go implementation with a nodejs one. None of the drawbacks of the current implementation would be addressed.

I think we can safely remove this bit.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
nelsonkopliku: +1 regardless of the implementation, whether the current external artifact or the embedded one, when connecting to trento's mcp server the user needs to be identified, in order to authorize operations.

I would imagine that a user still needs to configure the MCP client by providing the MCP server url and some additional config, where I imagine the PAT would still be a thing. 

What are the main concerns here? What am I missing? Is OAuth/OIDC involved in the topic? If so we're going to open a pandora box 🙈

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+== Security
+
+The Go mcp-server acts as a multi-user proxy between the client (your IDE or agent) and the Elixir trento-web API. The job of the MCP is to fetch OpenAPI specs.
+
+* *Token Delivery*: The MCP client (e.g., VS Code or Claude Desktop) is configured with an Authorization header containing the user's Trento Personal Access Token (PAT).
+* *Session Mapping*: When the client initializes the connection (via SSE/HTTP), the Go mcp-server intercepts the header, associates the PAT with the unique sessionID, and stores it.
+* *Tool Call Execution & Auth forwarding*: When a tool is executed, the Go server locks the execution context, injects the PAT into the process-global BEARER_TOKEN environment variable, runs the tool (which makes an HTTP request to trento-web), and restores the environment.
```
nelsonkopliku, line 59: **thought:** not sure how relevant the detail about the current implementation is here.

I mean what is relevant to me is that:
- a user connecting to the MCP server via a client needs to be authnz-ed
- a PAT is the way we have to let third party apps to call Trento APIs
- regardless of the implementation when connecting to trento's mcp server the user needs to be identified, and the PAT is the only/best option we currently have

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+== Configuration options
+
+You can create an `mcp-server-trento` configuration file in the working directory or specify a custom path using the `--config` flag.
+
+The Trento MCP Server searches for configuration files in the following order:
+
+1. System-wide directory (`/etc/trento/mcp-server-trento` or `/usr/etc/trento/mcp-server-trento`).
+2. Custom path specified with the `--config` flag.
+
+ NOTE: SLES 16 uses UsrEtc as the configuration scheme for packages. In UsrEtc, packages store the configuration by default as a .conf file or directory in /usr, and the actual configuration is either in a subdirectory of /etc/ or a .conf file in /etc. The configuration file should thus be upgraded to be either mcp-server-trento.conf or mcp-server-trento.d/ to comply with the UsrEtc scheme. Ideally, the configuration should be in /etc/trento/mcp-server-trento.conf.
```
nelsonkopliku, line 72: **thought:** this detail about how the current mcp server implementation handles configuration would fit better as part of its documentation and have here a reference to it, if deemed necessary.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+|VERBOSITY	|Kept, different form	|Elixir Logger level via config, not a -v flag.
+|===
+
+Not every option is superfluous, but the configuration is simpler and more aligned with the rest of Trento's configuration. The only options that remain are:
+
+[cols="3,1,6", options="header"]
+|===
+|Environment variable | default value | Detail
+| TRENTO_MCP_ENABLED | false | Whether the internal MCP server is enabled or not. If false, the internal MCP server will not be started.
+| TRENTO_MCP_PORT | 5000 | The port on which the MCP server listens for incoming connections with HTTP transport
```
nelsonkopliku, line 101: **issue:** this sounds in contradiction with what mentioned a few lines earlier

> PORT	|Merged	| For HTTP transport it's just the existing Phoenix endpoint port + a mounted route, not a separate listener.
>

> TRENTO_MCP_PORT | 5000 | The port on which the MCP server listens for incoming connections with HTTP transport
>

We either end up in having another endpoint in elixir that listens to a different port, or we have a different route in the current endpoint.

AI assistant should have been more careful 🙈

content/rfc/0005-migrate-to-internal-mcp-server.adoc
nelsonkopliku: Activity log is not related to the mentioned log level here. I understand we are talking about application logs.

Besides this, in the context of an elixir embedded MCP server I would start by only relying on the global web logger and its configuration.

Yes, I believe we can configure multiple loggers and what not, it's just a matter of avoiding accidental complexity where we have more than enough inherent complexity 😅

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+* *Authentication plug* — a plug on the MCP scope that validates the Trento PAT/session before the request reaches the server. This replaces HEADER_NAME and is required for anything non-local.
+* *Tool definitions* — one Anubis component module per exposed operation. This is the real work that OAS_PATH + TAG_FILTER were doing automatically; here it's explicit and hand-written.
+
+
+[cols="3,1,6", options="header"]
+|===
+| Setting | Default | Notes 
+|`config :trento, :mcp_server, enabled` | `false` | Gates the supervised server + makes `/mcp` inert when off 
+| `ENABLE_MCP_SERVER` (runtime env) | `false` | Sets the above in `config/runtime.exs` |
+|===
```
nelsonkopliku, line 121: **thought:** most of this could be replaced with references to [anubis](https://github.com/zoedsoupe/anubis-mcp) documentation [here](https://anubis-mcp.hexdocs.pm/readme.html).
That's not the point, though.

What concerns me more is the following:

> * *Tool definitions* — one Anubis component module per exposed operation. This is the real work that OAS_PATH + TAG_FILTER were doing automatically; here it's explicit and hand-written.
>

I am afraid we might be underestimating the complexity/consequences involved.

Here's an image that depicts how I feel about this

Image

Let me elaborate more:

There are currently 25 tools exposed by web https://demo.trento-project.io/api/doc#/MCP and they assume, where applicable:
- authnz is triggered (yes we could attach relevant plugs to anubis)
- operation policies are taken into account (yes we currently do not expose operations as tools and in case we could attach relevant plugs to anubis)
- relevant activities are tracked in the activity log (yes we could attach relevant plugs to anubis)

all of these need to stay valid when embedding the mcp server in trento web.

I am sure there are many options:
- we could hand-write all the anubis components at the cost of code duplication (spec + controller action implementation) and register relevant plugs where needed
- we could improve the `ai_tool` macro so that anubis component registration is automated
- we could further refactor/improve the codebase to better decouple service layer/business logic from edge entry points (http controllers, mcp server) so that plugging the same things together in different places is more ergonimic

Also none of the concerns about also wanda currently exposing tools via mcp seem to be considered 😅 

Bottom line is that nothing comes from free and that additionally to this RFC we might need some degree of iteration over concrete POCs to better evaluate options.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+
+=== Caveats
+
+1. The Go bridge that we were using before gives you the API for free, only taking into account the tags required to expose it. Using Anubis in Elixir is already automated. The same "MCP" tags used in the Trento server are used to configure the AI assistant's toolset and so that they continue being aligned automatically. The only new work is the authentication plug.
+
+2. Additional work needs to be done to add a health check endpoint to the internal MCP server, as the current health check is done by the Go bridge. The health check should be implemented in Elixir, reusing the existing liveness/readiness endpoints.
```
nelsonkopliku, line 128: **question:** besides the fact that removing the extra artifact clearly removes also all it involves, in the context of an embedded mcp server in trento web, what would the existing `/healthz` and `/readyz` endpoints need to do differently?

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+3. There is no way to install the MCP server as a separate application. By design, the internal MCP server is part of the Trento application, and it is not possible to install it separately.
```
nelsonkopliku, line 130: **nit:** not necessary content IMHO, the whole RFC _is about removing a separate application_.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
```diff
@@ -0,0 +1,140 @@
...
+
+4. You can't use command-line flags to configure the internal MCP server. The configuration is done via the Elixir config files or environment variables. Additional UX design should be done to include them as part of the command line for Trento itself.
```
nelsonkopliku, line 132: **nit:** not necessary content IMHO. The what and the how about configuration is an implementation detail at this stage.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
nelsonkopliku: +1 in deferring further discussions about umbrella application.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
nelsonkopliku: I also noticed the same inconsistency and left a similar related comment elsewhere.

TLDR: I don't see, yet, a need to have a different endpoint (as in the `Phoenix.Endpoint` meaning) for the embedded mcp server. It could very well be a route in the current endpoint.

Unless I am missing something.

nelsonkopliku (commented):
Thanks @chargio for the effort.

I am assuming that this RFC is coming from a clarification of the fact that from a broader product perspective we **DO NOT NEED** a separate artifact called `mcp-server-trento`

If that is the case then TLDR: 

I think the idea of simplifying the overall system by embedding the mcp server inside trento-web would have several benefits:
- for customers as they would have less operational burden
- for us as we would improve maintainability
- for the feature stability as it would reduce the networking roundtrip needed with the current proxy

However, as mentioned in the long-ish comment inline I believe this is a delicate decision that requires even more careful thinking and iteration on actual options.

If there was an expectation for this to happen before trento 3.2, I am afraid we will need to revisit that expectation.

Happy to discuss more, either in this RFC and/or sync.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: Clarified the text and added a link to kit, as mcphost is deprecated and substituted by it and it will be updated in SLES

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: Resolving this as the problem is not specific to the solution proposed

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: I've updated this part,

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: Adding an additional consideration to decide if we map the current log level to a new log level configuration or we merge it with the current log level.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: I understand that if the RFC is approved, we will need to work to refine the implementation.
Whether is done in 3.2 or later is a question of priorities and roadmap. 
This adds the requirement to the backlog

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: The current MCP requires configuration (you provide the port and the log level) and a bunch of other configurations in the mcp-server-trento.example.

I understand that supporting MCP server configuration to substitute MCP server configurations for the old server is just standard practice and does not require further explanation.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: Currently the endpoints are related to the endpoints in a different port. We publish endpoints for healthchecks at <trento>/mcp-server_endpoint/mcp /healthz and /readyz.
If we decide to update the endpoint to /mcp reusing Trento web endpoint, the healthz and readyz endpoints will be those of web, so we need to publish/develop new ones at /mcp/healthz and /mcp/readyz or drop them completely.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: I agree that the whole RFC is about removing it as a separate application, but I am highlighting that point. If the expectation is that it can be run as a separate application, it is not possible unless we evolve the architecture to support an umbrella app.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
chargio: This is a list of things that will require additional design or that can become a problem, in the idea that are not. Listing things that can be considered is not defining an implementation

chargio:
I had a strange behaviour where 3 files that were not touched by my PR were failing because of differences...
I've rebased and got again a weird behaviour.

content/rfc/0005-migrate-to-internal-mcp-server.adoc
nelsonkopliku, line 1: **issue:** this should be `0006-...` and we need to add an entry after https://github.com/trento-project/docs/blob/main/trento-docs-site/modules/developer/nav_developer.adoc?plain=1#L53

PR #260, opened 2026-08-05
Update versions for July patch release

eromanova97:
### Description
Update versions after patch release

Fixes https://jira.suse.com/browse/TRNT-4499

### Preview
image


trento/adoc/trento-version-history.adoc
antgamdia: Maybe this needs a change as well?

antgamdia (approved):

trento/adoc/trento-version-history.adoc
eromanova97: I would almost forget, thanks!

PR #267, opened 2026-08-12
Initial definition of a PR policy

chargio:
### Description

Add  a description of the required changes in documentation to address the needs for better PR requests and reviews.
This document describes the changes that will need to be implemented.

Fixes #<issue>

### How was this tested?

Document only

### Documentation changes

Yes


skrech (changes requested):
I refuse to accept this AI-generated document. Overall, the target audience is unclear -- is it written for our team, is it intended for external contributors to read? 

It's an enormous amount of text that re-iterates the same common-sense practices. 

It is clear to me that separate sections were changed several times without the context of the wider document. 

There are refenreces that are not defined -- what is "CL"? 

The granularity of the suggested changes is inconsistent -- some recommendations are very vague while others are super detailed -- for example, setting up of stalebot. Ironically, a clear sign that this is written with AI, not complying with the same rules that this document is trying to establish.

chargio:
Thanks for the feedback @skrech.

I will have a review on it to improve the style. Just to clarify, it is not AI generated (although I've used AI extensively). It just covers a lot of content from different sources, and that can be why you find differences in details. There are best practices that are common  and well established, while others are not.

Apart from the style, do you have any comment on the content?

content/rfc/0007-updated-pr-policy.adoc
```diff
@@ -0,0 +1,259 @@
...
+2. *Zero tolerance Triage Rule*: Maintainers will immediately close - _without review of explanations_ - any PR exhibiting standard AI hallucinations (non-existing API calls, dead imports, phantom config flags, or generic boilerplate code) or any other code that does not meet the project's quality standards.
+3. *Code contributions will be accompanied by a clear explanation of the code's purpose, functionality, and any relevant context*. You should not expect reviewers to guess the rationale for the code. This will help maintainers understand the code and its intended use, and will facilitate the review process. _PRs that do not provide sufficient context or explanation may be closed without review_.
+4. *Code contributions will express in writing in the PR commit that they followed all project guidelines, including code style, testing, and documentation requirements*. This will help ensure that all contributions meet the project's quality standards and are consistent with the existing codebase. PRs that do not include this statement may be closed without review.
+
+=== How Fast should code reviews be?
```
skrech: I disagree with the suggestion for such fast responses. The response on a PR is expected to be an actual review, which necessitates a holistic overview. Reviews can't be in a chat form. Depending on the change this might involve a lot of thinking, if you're occupied with feature-development work that is not going according to plan or involves intense thinking, it's hard to meet that expectations. Even having a limit is intimidating. What if we don't review in the specified time-frame? I think we need to use more prescriptive language, not enforcing one.

chargio: Thanks. That makes sense. The fast response is explicit in the Google document that we use today as a reference

content/rfc/0007-updated-pr-policy.adoc
```diff
@@ -0,0 +1,259 @@
...
+
+
+== Updated Pull Request Workflow (for reviewers)
+
+=== Step 0: Pre-work
+Before a PR is submitted, the developer needs to have a clear understaning of the scope and requirements. If refinement work is needed, consider both including the refinement documentation as context in the PR and choosing reviewers that have participated in the refinement and design of the solution. For that reason, it is important to involve the relevant stakeholders early in the process to ensure that the PR is well-informed and aligned with the project's goals.
+
+=== Step 1: Initial assignment
+The PR can be assigned to a squad (preferably) or to one or several individuals, or left unassigned for maintainers to pick up. In the last case where it is left unassigned, maintainers will review the PR and assign it to the appropriate person or team for further review and action. Unless it is specifically requested, the PR needs to have one successful maintainer review before merging and no rejections, so assigning it to a squad will generally help streamline this process. AI approval does not replace the requirement for a human maintainer review.
```
skrech: Who is the `maintainer` here? It seems that this designates some form of code-ownership, but we don't have such thing.

content/rfc/0007-updated-pr-policy.adoc
```diff
@@ -0,0 +1,259 @@
...
+
+=== Step 2: Take a broad view of the change
+Look at the PR description and what the PR does in general. Does this change even make sense? If this change shouldn't have happened in the first place, please respond immediately with an explanation and why the change should not be happening. It is a good idea to suggest to the developer what they should have done instead.
+
+If the code is not understandable, it is clearly not ready for review or it manifests AI hallucinations, it should be immediately flagged and the PR should be closed. The developer can open a new pull request after addressing the issues.
+
+=== Step 3: Pre-review automation gate
+Once step 1 is done, the reviewer will include an AI-assisted verification step to automatically check for code quality, maintainability, and security issues. This step will help ensure that only high-quality contributions proceed to the manual review stage. Step 1 will ensure that AI is not engaged inappropriately or without proper oversight.
```
skrech: This step could potentially help, but we see that AI reviews include a lot of false-positives but at the same time lengthy explained. A prudent reviewer would probably go over these as well... which only increases the burden, not lower it. 

AI-assisted reviews should only be done on-demand, if the reviewer wants to be helped by a pre-flight ai check, or the contributor wants to do that locally before PR submission.

antgamdia (changes requested):
First of all, thank you for the effort in putting all this together. 

In general, while I agree on the need of a PR review policy in the team, I'm also pretty reluctant to +1ing this one. There are interesting parts, for sure, but it is inconsistently verbose, and I find rather difficult to extract what is really important from usual best practices. 

For example, we have a section on "merge strategies", when it can be enforced from the repository settings in a checkbox.  

I miss more instructions on PR title, labeling, tracebility with a ticket (whether it be an internal one in Jira or a public one if a user opens an issue), etc...

There's some minor changes requested, like not really updating the CONTRIBUTING.md or other unrelated files in this very PR.

Honestly, while I value the concept of an RFC publicly available, I feel this document should have been a more agile one first (gDocs for example), so reviewing it is not that static or "serious". Once we all kind of agree in the content, I would convert it into an RFC, where only the format and, perhaps, the style, are in stake; but not the teleological intention and overall content.

content/rfc/0007-updated-pr-policy.adoc
```diff
@@ -0,0 +1,259 @@
...
+
+=== Step 4: Manual review
+After the automatic verification step is successfully completed, the reviewer will manually review the code changes. This includes checking for correctness, adherence to project guidelines, code readability, and overall quality. The reviewer will provide feedback, request changes if necessary, and approve the pull request once it meets all standards.
+During this phase, AI-assisted verification tools may still be used to support the review process, but the final judgment and approval rest with the human reviewer.
+
+When the AI identifies the PR as low risk and trivial, and if the CI is passing, the reviewer may choose to expedite the approval process, potentially merging the PR with minimal manual intervention. In order to do so, the reviewer should document the rationale for the expedited approval in the PR comments, ensuring transparency and traceability for future reference. If the merge of the PR imples that the build is broken, additional CI tests will need to be put in place to avoid errors in the future.
```
skrech: I feel the goal if this RFC is to try to mask a push for integrating AI as review process establishment/improvement. It feels that integrating AI into the review process is the sole purpose of the document. A goal in itself.

The part about "PR imples that the build is broken" is totally out of space -- all checks are happening before merge, not after that.

content/rfc/0007-updated-pr-policy.adoc
```diff
@@ -0,0 +1,259 @@
...
+
+Reviewers should always feel free to leave comments expressing that something could be better, but if it’s not very important, prefix it with something like “Nit: “ to let the author know that it’s just a point of polish that they could choose to ignore.
+
+== Pre-Review Automation Gate
+To avoid unnecessary maintainer bandwidth consumption, a pre-review automation gate will be implemented to automatically check for code quality, maintainability, and security issues before a pull request is assigned to a maintainer for review. This gate will help ensure that only high-quality contributions are reviewed and merged into the codebase.
+
+1. *Manual Trigger*: To avoid unnecessary CI/CD costs, this gate will be triggered manually by a maintainer after the initial triage.
```
skrech: There is no reason for this check to be manually triggered. Too vague description here -- what code-quality? What maintainability? 
Overall, what is the gate? Github check? Or CI check?

content/rfc/0007-updated-pr-policy.adoc
```diff
@@ -0,0 +1,259 @@
...
+2. *Maintainers won't perform human reviews until 100% of automated checks pass*: Once the automation gate is passed, the pull requests will have passed all automated checks (linter, tests, commit linting, PR size, PR template completion, etc).
+3. *AI-best practices configuration*: The project will include guidelines and configurations to ensure that AI-assisted contributions adhere to best practices, including proper disclosure, verification, and quality standards. Including architecture documents for AI, AI skills and other relevant documentation will be part of this configuration.
+
+== Component changes
+
+=== Addition of CONTRIBUTING.adoc file
+The contribution guide will be enhanced with the following sections:
+Scope and workflow:
+- **Issue first**: PR for new features or structural changes must be linked to an existing issue. PRs without an issue will be closed without review.
```
skrech: This is unapplicable for our case. We have bug-tracker that is private. We should not reference private bug trackers. If the contribution is from external contributor, then this might be viable, but in my opinion, it's not necessary. Good PR description for a focused change is enough.
However, if there is a Issue opened for a problem/feature and there is a fix for that, then the PR should reference that issue.

chargio:
Closing PR to move the conversation to a more suitable place, as suggested

PR #268, opened 2026-08-13
Update PR env ssl cert renewal

nelsonkopliku:
### Description

Update the guide to renew PR env ssl cert so both https://prenv.trento.suse.com/ and https://foo.prenv.trento.suse.com/ actually have a proper certificate.

### DIFF
--- a/content/internal-notes/pr-env-ssl-certificate-setup.adoc
+++ b/content/internal-notes/pr-env-ssl-certificate-setup.adoc
@@ -13,30 +13,61 @@ Run the following command to start the certificate generation process:
 
 [source,bash]
 ----
-certbot certonly --manual --preferred-challenges dns -d "*.<the-domain>"
+certbot certonly --manual --preferred-challenges dns \
+  --cert-name <the-domain> \
+  -d "*.<the-domain>" \
+  -d "<the-domain>"
 ----
+
+Both `+-d+` flags are required. A wildcard matches exactly one label and
+does not cover the bare domain it is derived from, so a certificate
+issued only for `+*.<the-domain>+` is valid for `+1234.<the-domain>+`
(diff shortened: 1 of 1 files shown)


gagandeepb (approved):
Thanks for this.

content/internal-notes/pr-env-ssl-certificate-setup.adoc
```diff
@@ -13,30 +13,61 @@ Run the following command to start the certificate generation process:
...
 
-Before continuing, verify the record is deployed.
+Before continuing, verify the TXT record has been deployed.
 ....
 
+The two names are validated separately, but both challenges use the
+*same* record name, `+_acme-challenge.<the-domain>+`, with different
+values. Certbot prints them one at a time:
+
+* If it prompts *twice*, the second value must be *added alongside* the
```
arbulu89, line 51: question:
"prompts twice", you mean it prints 2 lines with two strings?
Or it does the whole sequence twice?

arbulu89 (approved):
No issues with this.
The other option was to disable the "domain only" serving, so it doesn't even exist.
We don't really use it.

PD: I think at some point we need to spend some time and enable the auto renew of certificates if possible

content/internal-notes/pr-env-ssl-certificate-setup.adoc
nelsonkopliku: first code -> route53 -> enter -> then second code and again

(I actually falled in the once scenario only because I had renewed brfore 😅 )

arbulu89: Ok, the whole process of getting code and putting in route53 (and waiting) twice (second time appending the value).
got it

nelsonkopliku:
@arbulu89 big +1 in automating the renewal. Let me add an item in the backlog

